PR #6, opened 2024-11-10
feat: add `no-invalid-hex-color` css rule

akulsr0:
#### Prerequisites checklist

Add `no-invalid-hex-color` css rule

Add `no-invalid-hex-color` css rule, which warns when it finds an invalid hex color in your css code.


nzakas:
Thanks for the submission. In general, we like to have people open issues with rule suggestions before implementing them. That allows us to have a conversation before anyone puts in too much work. 

In this case, I believe CSSTree has a way to flag not just [invalid colors](https://github.com/csstree/csstree/blob/bd4668fe843006387fccfd9786ecc4e0af384c68/lib/lexer/generic.js#L246) but also invalid values, so I'm not sure it makes sense to have such a narrowly-defined rule that just checks hex colors when we really want to validate all kinds of values. So, I think we need to take a step back here and think about what it means to validate more than one type of value.

PR #9, opened 2024-11-15
feat: no-invalid-property-values rule

nzakas:
#### Prerequisites checklist

Add a rule to validate property values. This uses the CSS spec to check every value of every property to ensure it's correct and expected.

- Added `no-invalid-property-values` rule
- Added tests
- Added docs
- Updated README
- Added `no-invalid-property-values` rule to `recommended` config

#### Related Issues

Refs https://github.com/eslint/css/pull/6

I'm a bit torn as to whether to combine this rule with `no-unknown-properties`, because `lexer.matchDeclaration` will also check for unknown properties. I'd like some opinions on this. Here's what that looks like:

https://github.com/eslint/css/pull/11

cc @akulsr0

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,12 +56,13 @@ export default [
 
 <!-- Rule Table Start -->
 
-| **Rule Name**                                                    | **Description**                   | **Recommended** |
-| :--------------------------------------------------------------- | :-------------------------------- | :-------------: |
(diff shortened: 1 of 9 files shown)


nzakas:
The types for the `css-tree` package are messed up and causing the build to fail even though the code runs fine. I'll address that next week.

akulsr0:
@nzakas I would prefer it to be not combined with `no-unknown-properties` rule. Keeping both rules easily configurable independently.

nzakas:
Closing in favor of #11.

PR #23, opened 2024-11-27
feat: add no-important rule

yannbertrand:
#### Prerequisites checklist

Add a rule to avoid `!important` annotation usage.

- Created the rule
- Added it to the recommended rules
- Wrote the docs

#### Related Issues

See #20

Implementation was not complicated but I'm not sure on how much details the docs should include.

(It's my first PR on ESlint, lmk if anything is wrong or if the rule is useless)

### DIFF
--- a/docs/rules/no-important.md
+++ b/docs/rules/no-important.md
@@ -1,0 +1,21 @@
+# no-important
+
+Disallow `!important` annotations.
+
+## Background
+
+Needing !important indicates there may be a larger underlying issue.
+
+## Rule Details
+
+Examples of incorrect code:
+
+```css
+a {
+	color: red !important;
+}
+
+a .link {
+	font-size: padding: 10px 20px 30px 40px !important;
+}
+```
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ export default [
 | :--------------------------------------------------------------- | :------------------------------- | :-------------: |
 | [`no-duplicate-imports`](./docs/rules/no-duplicate-imports.md)   | Disallow duplicate @import rules |       yes       |
 | [`no-empty-blocks`](./docs/rules/no-empty-blocks.md)             | Disallow empty blocks            |       yes       |
+| [`no-important`](./docs/rules/no-important.md)                   | Disallow !important annotations  |       yes       |
 | [`no-invalid-at-rules`](./docs/rules/no-invalid-at-rules.md)     | Disallow invalid at-rules        |       yes       |
 | [`no-invalid-properties`](./docs/rules/no-invalid-properties.md) | Disallow invalid properties      |       yes       |
 


src/rules/no-important.js
```diff
@@ -0,0 +1,41 @@
...
+//-----------------------------------------------------------------------------
+// Rule Definition
+//-----------------------------------------------------------------------------
+
+export default {
+	meta: {
+		type: /** @type {const} */ ("problem"),
+
+		docs: {
+			description: "Disallow important annotations",
```
nzakas, line 15:
```suggestion
			description: "Disallow !important annotations",
```

src/rules/no-important.js
```diff
@@ -0,0 +1,41 @@
...
+			recommended: true,
+		},
+
+		messages: {
+			forbiddenImportantUsage: "!important annotations are not allowed.",
```
nzakas, line 20: Can we shorten this to `unexpectedImportant`?

And we generally like to be a bit gentler with the messages:

```suggestion
			unexpectedImportant: "Unexpected !important annotation found.",
```

src/rules/no-important.js
```diff
@@ -0,0 +1,41 @@
...
+
+	create(context) {
+		return {
+			"Declaration[important=true]"(node) {
+				context.report({
+					loc: {
+						start: node.loc.start,
+						end: {
+							line: node.loc.start.line,
+							column:
```
nzakas, line 32: This will just highlight the property name. I think it would be better to find the location of the `!important` and set that as the location of the error.

tests/rules/no-important.test.js
```diff
@@ -0,0 +1,89 @@
...
+	},
+	language: "css/css",
+});
+
+ruleTester.run("no-important", rule, {
+	valid: [
+		"a { color: red; }",
+		"a { color: red; background-color: blue; }",
+		"a { color: red; transition: none; }",
+		"body { --custom-property: red; }",
```
nzakas, line 30: I'd also add a test with just the word "important":

```suggestion
		"body { --custom-property: red; }",
		"body { --custom-property: important; }",		
```

tests/rules/no-important.test.js
```diff
@@ -0,0 +1,89 @@
...
+		{
+			code: "a { color: red !important; }",
+			errors: [
+				{
+					messageId: "forbiddenImportantUsage",
+					data: {
+						property: "color",
+						value: "red !important",
+						expected: "red",
+					},
```
nzakas, line 46: `data` doesn't apply because you're not passing any `data` to `context.report()` in the rule. Please remove all `data` entries from this file.

tests/rules/no-important.test.js
```diff
@@ -0,0 +1,89 @@
...
+					line: 1,
+					column: 11,
+					endLine: 1,
+					endColumn: 18,
+				},
+			],
+		},
+	],
+});
```
nzakas, line 89: Can you also add a test that has multiple properties with `!important` to ensure we're catching them all?

nzakas (changes requested):
Thanks for getting this started, it looks really good. Just a few things to clean up.

docs/rules/no-important.md
```diff
@@ -0,0 +1,21 @@
+# no-important
+
+Disallow `!important` annotations.
+
+## Background
+
+Needing !important indicates there may be a larger underlying issue.
```
nzakas, line 7: We need more of a description here. :smile: Please see other rule docs for examples.

docs/rules/no-important.md
```diff
@@ -0,0 +1,21 @@
...
+
+```css
+a {
+	color: red !important;
+}
+
+a .link {
+	font-size: padding: 10px 20px 30px 40px !important;
+}
+```
```
nzakas, line 21: Please also add "Prior Art" section as in other rules.

https://stylelint.io/user-guide/rules/declaration-no-important/
https://github.com/CSSLint/csslint/wiki/Disallow-%21important

src/rules/no-important.js
yannbertrand: Done

src/rules/no-important.js
yannbertrand: Done. Thanks for the suggestions

tests/rules/no-important.test.js
yannbertrand: Missed that one 😅 thanks, it's fixed

yannbertrand:
> Thanks for your work on this. There's some ongoing discussion about whether or not we'd like to have this rule on #20, so we may need to think things through a bit before proceeding.

Sure thing, no problem! Should keep this PR opened? I'll wait for a decision to either get back to work on this one or replace it by another PR depending on the expected scope.

nzakas:
Let's keep this one open for now. I'll just mark it as a draft while we try to figure things out.

nzakas:
@yannbertrand sorry for the delay on this. We'd like to move forward. If you can please take a look at resolving the conflicts and addressing the feedback that was left, we'd appreciate it.

nzakas:
@thecalamiity yes, absolutely. That would be appreciated. :pray:

PR #24, opened 2024-11-27
feat: max-specificity rule

### LINKED ISSUE #19: New Rule: selector-complexity
### Rule details

Flag any CSS selector that uses the ID selector

### Example code

```css
#one { /* lint error */
  background-color: yellow;
}

h1#heading { /* lint error */
  color: rebeccapurple;
}
```

The `#id` selector has very high specificity and doesn't always play well with other rules. An alternative is to use a `.class` selector instead.

When using a component-based UI framework, id selectors are often not very useful. You can hardcode an id into a component, but this makes the component non-reusable. The framework-generated unique ids cannot be referenced in CSS. Therefore, using id selectors in these cases is likely a mistake.

### Participation

### Additional comments

[From MDN:](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Type_Class_and_ID_Selectors#id_selectors)

> The ID selector has high specificity. This means styles applied based on matching an ID selector will overrule styles applied based on other selector, including class and type selectors. Because an ID can only occur once on a page and because of the high specificity of ID selectors, it is preferable to add a class to an element instead of an ID. If using the ID is the only way to target the element — perhaps because you do not have access to the markup and cannot edit it — consider using the ID within an attribute selector, such as p[id="header"].

nzakas:
#### Prerequisites checklist

Implement a `max-specificity` rule.

- Created `max-specificity` rule
- Added tests
- Added to plugin exports

#### Related Issues

fixes #19

This is a rough draft, I'd like some feedback on the design.

- Does it make sense to have people write out `a`, `b`, and `c` to match the spec? Or would it better to just have them enter an array?
- Would any other options make sense, in which case, the desired specificity should be one key in an options object?

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,12 +56,13 @@ export default [
 
 <!-- Rule Table Start -->
 
-| **Rule Name**                                                    | **Description**                  | **Recommended** |
-| :--------------------------------------------------------------- | :------------------------------- | :-------------: |
-| [`no-duplicate-imports`](./docs/rules/no-duplicate-imports.md)   | Disallow duplicate @import rules |       yes       |
-| [`no-empty-blocks`](./docs/rules/no-empty-blocks.md)             | Disallow empty blocks            |       yes       |
-| [`no-invalid-at-rules`](./docs/rules/no-invalid-at-rules.md)     | Disallow invalid at-rules        |       yes       |
-| [`no-invalid-properties`](./docs/rules/no-invalid-properties.md) | Disallow invalid properties      |       yes       |
+| **Rule Name**                                                    | **Description**                                  | **Recommended** |
+| :--------------------------------------------------------------- | :----------------------------------------------- | :-------------: |
+| [`max-specificity`](./docs/rules/max-specificity.md)             | Enforce the maximum specificity of CSS selectors |       no        |
+| [`no-duplicate-imports`](./docs/rules/no-duplicate-imports.md)   | Disallow duplicate @import rules                 |       yes       |
+| [`no-empty-blocks`](./docs/rules/no-empty-blocks.md)             | Disallow empty blocks                            |       yes       |
+| [`no-invalid-at-rules`](./docs/rules/no-invalid-at-rules.md)     | Disallow invalid at-rules                        |       yes       |
+| [`no-invalid-properties`](./docs/rules/no-invalid-properties.md) | Disallow invalid properties                      |       yes       |
 
 <!-- Rule Table End -->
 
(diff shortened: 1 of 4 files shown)


jeddy3:
We added [`selector-max-specificity`](https://stylelint.io/user-guide/rules/selector-max-specificity/) to Stylelint before the introduction of specificity-changing pseudo-classes like [`:is()`](https://drafts.csswg.org/selectors/#matches), [`:where()`](https://drafts.csswg.org/selectors/#zero-matches), [`:not()`](https://drafts.csswg.org/selectors/#negation) and [`:has()`](https://drafts.csswg.org/selectors/#relational). Having thought about it some more, I'm not sure the rule is useful these days.

Due to these pseudo-classes, it makes a poor proxy for controlling selector complexity and for disallowing ID selectors. For example, both these selectors have a specificity of `0,0,0`:

```css
:where(#id),
:where(a.complex[attr], #id) {}
```

And makes a poorer proxy for disallowing other types of selectors as it [groups them](https://drafts.csswg.org/selectors/#specificity-rules):

- B = the number of class selectors, attributes selectors, and pseudo-classes
- C = the number of type selectors and pseudo-elements

If you keep the rule, the implementation should probably consider these specificity-changing pseudo-classes and [nesting](https://drafts.csswg.org/css-nesting/), which is [Baseline 2023](https://caniuse.com/css-nesting).

In Stylelint, we use [`@csstools/selector-specificity`](https://github.com/csstools/postcss-plugins/blob/main/packages/selector-specificity/src/index.ts) for calculating the specificity, which supports the specificity-changing pseudo-classes. We also desugar nested selectors before passing them to the calculator. However, we don't yet [desugar to `:is()`](https://drafts.csswg.org/css-nesting/#nest-selector) (as our current implementation predates the latest CSS nesting spec):

```css
a, b { 
  & c {} 
}

/* should be desugared to */
:is(a, b) c {}

/* be we currently desugar to */
a c, b c {}
```

This can lead to [different specificities](https://drafts.csswg.org/css-nesting/#example-a71fa710), and our calculations may not match the browser.

nzakas:
Interesting. I think I'll need to dig into this a bit more.

lahmatiy:
Take a look at [bramus/specificity](https://github.com/bramus/specificity). It appears to to cover all the key aspects of selector specificity calculation according to the latest CSS specifications. Since it’s built on top of CSSTree, it might be a good candidate for adoption instead of implementing a custom specificity calculation from scratch (see [calculate.js](https://github.com/bramus/specificity/blob/main/src/core/calculate.js)).

nzakas:
@lahmatiy thanks! We're actually discussing over on #19 whether or not specificity even makes sense as a measure of selector complexity anymore. Would love to hear your thoughts.

nzakas:
Closing for now as this doesn't seem like the right approach.

PR #27, opened 2024-12-04
feat: Add use-layers rule

### LINKED ISSUE #18: New Rule: use-layers
### Rule details

Flag any CSS rules that are outside a `@layer` block.

### Example code

This will be flagged:

```css
/* ❌ */

p {
  color: red;
}

@media (width > 500px) {
  .box {
    padding: 12px;
  }
}
```

This is fine:

```css
/* ✅ */

@layer foo {
  p {
    color: red;
  }
}

@media (width > 500px) {
  @layer bar {
    .box {
      @layer baz {
        padding: 12px;
      }
    }
  }
}
```

Note: Nested layers can used as `@layer foo.bar {…}` and anonymous layers can be used as `@layer {…}`. Both of these are fine and should not be flagged.

Some CSS, like `@import`, is fine to keep outside a `@layer`, because the contents of the file would still be using `@layer`.

```css
@import "foo.css";
@import "bootstrap.css" layer(thirdparty);

@layer foo {…}
```

### Participation

- [ ] I am willing to submit a pull request to implement this rule.

### Additional comments

When working with [cascade layers](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_layers#creating_cascade_layers), unlayered styles are usually a mistake. It is easy to forget to wrap the styles in `@layer`, which is exactly the kind of thing a linter should help with.

This rule probably should be opt-in, because cascade layers aren't being widely used yet.

nzakas:
#### Prerequisites checklist

Add a new `use-layers` rule

- Added `use-layers` rule
- Added tests
- Added docs
- Ensured the rule is exported from the plugin

Note: This rule is *not* recommended because it will likely only be useful on a subset of CSS files in any given project.

#### Related Issues

fixes #18 

- ~Should `@import` require a layer by default? Right now it doesn't, but I feel like maybe it should.~ Yes!
- CSSTree is returning the wrong location for layer names. I've opened an issue for that: https://github.com/csstree/csstree/issues/309
-


jeddy3:
> Should `@import` require a layer by default? Right now it doesn't, but I feel like maybe it should.

I think so. I had [the same thought](https://github.com/eslint/css/issues/18#issuecomment-2518102505) before seeing this PR.

nzakas:
@fasttime I can't quite figure out why the type tests are failing now. Can you take a look?

nzakas:
@jeddy3 thanks, updated!

nzakas:
I think I figured out the types issue: https://github.com/eslint/css/pull/28

src/rules/use-layers.js
```diff
@@ -0,0 +1,138 @@
...
+					},
+					layerNamePattern: {
+						type: "string",
+					},
+				},
+				additionalProperties: false,
+			},
+			minItems: 0,
+			maxItems: 1,
+		},
```
fasttime, line 38: I think the `schema` property should be an array to match the `JSONSchema4` types. That's how most of our rules are defined.

```suggestion
		schema: [
			{
				type: "array",
				items: {
					type: "object",
					properties: {
						allowUnnamedLayers: {
							type: "boolean",
						},
						requireImportLayers: {
							type: "boolean",
						},
						layerNamePattern: {
							type: "string",
						},
					},
					additionalProperties: false,
				},
				minItems: 0,
				maxItems: 1,
			},
		],
```

The TypeScript error message indicates the problem (somehow) here:

          The types of 'meta.schema' are incompatible between these types.
            Type '{ type: string; items: { type: string; properties: { allowUnnamedLayers: { type: string; }; requireImportLayers: { type: string; }; layerNamePattern: { type: string; }; }; additionalProperties: boolean; }; minItems: number; maxItems: number; }' is not assignable to type 'false | JSONSchema4 | JSONSchema4[] | undefined'.

It's unobvious, but `schema` needs to be an array to satisfy `JSONSchema4` (unless it's defined with `anyOf`, `oneOf`, etc.).

nzakas: Ah gotcha, thanks!

nzakas:
I fixed the bug in CSS Tree with layer positioning. Just waiting for a new release to update this PR.

tests/rules/use-layers.test.js
```diff
@@ -0,0 +1,263 @@
...
+					@layer bar {
+						b { color: red; }
+					}
+				}
+			}
+		`,
+		{
+			code: "@layer { a { color: red; } }",
+			options: [{ allowUnnamedLayers: true }],
+		},
```
jeddy3, line 51: Worth adding tests for importing into unnamed layers? e.g.

```css
@import 'foo.css' layer;
@import 'foo.css' layer();
```

jeddy3 (commented):
I've added two suggestions for extra tests. 

I think I'd find it more intuitive if the boolean options matched, i.e. both made the rule stricter or more permissive.

For example, either:

- `allowUnnamedLayers: false`
- `allowUnLayeredImports: false`

or:

- `requireNamedLayers: true`
- `requireImportLayers: true`

tests/rules/use-layers.test.js
```diff
@@ -0,0 +1,263 @@
...
+			errors: [
+				{
+					messageId: "missingLayerName",
+					line: 1,
+					column: 19,
+					endLine: 1,
+					endColumn: 24,
+				},
+			],
+		},
```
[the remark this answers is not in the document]
nzakas: Good idea! I didn't realize that was possible.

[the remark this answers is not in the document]
nzakas: It turns out `layer()` is a parsing error, so we don't need a test.

tests/rules/use-layers.test.js
nzakas: For the first one, yes. As noted in my other comment, `layer()` is actually a syntax error.

nzakas:
With CSSTree updated, all of the locations are now correct. :tada:

src/rules/use-layers.js
```diff
@@ -0,0 +1,134 @@
...
+
+			Layer(node) {
+				if (!layerNameRegex) {
+					return;
+				}
+
+				/*
+				 * Layer node location is incorrect.
+				 * https://github.com/csstree/csstree/issues/309
+				 */
```
mdjermanovic, line 94:
```suggestion
```

This comment can be removed now since the problem with locations has been fixed?

mdjermanovic (approved):
LGTM, thanks!

PR #52, opened 2025-02-18
fix: require-baseline should not warn for cursor property

nzakas:
#### Prerequisites checklist

Update `require-baseline` rule does it doesn't warn for `cursor`.

- Updated `generate-baseline.js` to not include data for `cursor`.
- Added a test to the `require-baseline` rule to validate that `cursor` doesn't trigger a warning.
- Updated the `require-baseline` docs to highlight that we are deviating from baseline in this case.

#### Related Issues

refs #51


mdjermanovic (approved):
LGTM, thanks!

PR #61, opened 2025-02-22
feat: add selector support to require-baseline

### LINKED ISSUE #60: Bug: `:has()` not flagged with Baseline Widely available config
### Environment

ESLint version: 9.20.1
@eslint/css version: 0.4.0
Node version: 20.0.0
npm version: 9.6.4
Operating System: MacOS 14.7.3

stylesheet

### What did you do?

<details>

```
import globals from "globals";


/** @type {import('eslint').Linter.Config[]} */
export default [
  {
    languageOptions: {
      globals: globals.browser
    }
  },
  {
    files: ["**/*.css"],
    plugins: {
      css,
    },
    language: "css/css",
    rules: {
      "css/require-baseline": ["warn", {
        "available": "widely"
      }],
    },
  },
];
```
</details>

```css
/* baseline/has (low) */
p:has(i) {
  color: sandybrown;
}
```

The `:has()` pseudo-class is not yet Baseline Widely available, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:has):

Image

I would have expected `npx eslint src/has.css` to emit a warning that it "is not a widely available baseline feature"

### What actually happened?

No warnings were emitted

https://github.com/rviscomi/baseline-demos/tree/main/eslint

### Participation

### Additional comments

_No response_

rviscomi:
#### Prerequisites checklist

Add support for validating the Baseline status of selectors like `:has()`

- added a new `Selector` handler that checks the validity of all child nodes
- added tests for pseudo-class and pseudo-element selectors
- updated docs

#### Related Issues

Fixes #60

### DIFF
--- a/src/rules/require-baseline.js
+++ b/src/rules/require-baseline.js
@@ -15,6 +15,7 @@ import {
 	atRules,
 	mediaConditions,
 	types,
+	selectors,
 } from "../data/baseline-data.js";
 import { namedColors } from "../data/colors.js";
 
@@ -347,6 +348,8 @@ export default {
 				"Type '{{type}}' is not a {{availability}} available baseline feature.",
 			notBaselineMediaCondition:
 				"Media condition '{{condition}}' is not a {{availability}} available baseline feature.",
+			notBaselineSelector:
+				"Selector '{{selector}}' is not a {{availability}} available baseline feature.",
 		},
 	},
 
@@ -623,6 +626,48 @@ export default {
 							availability,
 						},
 					});
+				}
+			},
+
+			Selector(node) {
+				for (const child of node.children) {
+					const selector = child.name;
+
+					if (!selectors.has(selector)) {
+						continue;
+					}
+
+					const ruleLevel = selectors.get(selector);
+
+					if (ruleLevel < baselineLevel) {
+						const loc = child.loc;
+
+						// some selectors are prefixed with the : or :: symbols
+						let prefixSymbolLength = 0;
+						if (child.type === "PseudoClassSelector") {
+							prefixSymbolLength = 1;
+						} else if (child.type === "PseudoElementSelector") {
+							prefixSymbolLength = 2;
+						}
+
+						context.report({
+							loc: {
+								start: loc.start,
+								end: {
+									line: loc.start.line,
+									column:
+										loc.start.column +
+										selector.length +
+										prefixSymbolLength,
+								},
+							},
+							messageId: "notBaselineSelector",
+							data: {
+								selector,
+								availability,
+							},
+						});
+					}
 				}
 			},
 		};
--- a/src/rules/require-baseline.js
+++ b/src/rules/require-baseline.js
@@ -637,9 +637,9 @@ export default {
 						continue;
 					}
 
-					const ruleLevel = selectors.get(selector);
+					const selectorLevel = selectors.get(selector);
 
-					if (ruleLevel < baselineLevel) {
+					if (selectorLevel < baselineLevel) {
 						const loc = child.loc;
 
 						// some selectors are prefixed with the : or :: symbols
--- a/src/rules/require-baseline.js
+++ b/src/rules/require-baseline.js
@@ -125,6 +125,12 @@ class SupportsRule {
 	#properties = new Map();
 
 	/**
+	 * The selectors supported by this rule.
+	 * @type {Set<string>}
+	 */
+	#selectors = new Set();
+
+	/**
 	 * Adds a property to the rule.
 	 * @param {string} property The name of the property.
 	 * @returns {SupportedProperty} The supported property object.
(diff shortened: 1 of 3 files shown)


rviscomi:
I wasn't sure how to handle selectors used within `@supports`. It seems like the AST always parses them as raw text rather than `FeatureFunction` as suggested in the [docs](https://github.com/csstree/csstree/blob/master/CHANGELOG.md#300-september-11-2024).

Example CSS:

```css
@supports selector(.example) {}
```

AST:

```json
{
  "type": "Function",
  "loc": {
    "source": "<unknown>",
    "start": {
      "offset": 10,
      "line": 1,
      "column": 11
    },
    "end": {
      "offset": 28,
      "line": 1,
      "column": 29
    }
  },
  "name": "selector",
  "children": [
    {
      "type": "Raw",
      "loc": {
        "source": "<unknown>",
        "start": {
          "offset": 19,
          "line": 1,
          "column": 20
        },
        "end": {
          "offset": 27,
          "line": 1,
          "column": 28
        }
      },
      "value": ".example"
    }
  ]
}
```

https://astexplorer.net/#/gist/244e2fb4da940df52bf0f4b94277db44/ff8f2e9e7ca35af89743f13ae98f37a69e501e9d

src/rules/require-baseline.js
```diff
@@ -625,6 +628,48 @@ export default {
...
+					const selector = child.name;
+
+					if (!selectors.has(selector)) {
+						continue;
+					}
+
+					const ruleLevel = selectors.get(selector);
+
+					if (ruleLevel < baselineLevel) {
+						const loc = child.loc;
```
nzakas, line 643:
```suggestion
					const selectorLevel = selectors.get(selector);

					if (selectorLevel < baselineLevel) {
						const loc = child.loc;
```

src/rules/require-baseline.js
```diff
@@ -625,6 +628,48 @@ export default {
...
+
+			Selector(node) {
+				for (const child of node.children) {
+					const selector = child.name;
+
+					if (!selectors.has(selector)) {
+						continue;
+					}
+
+					const ruleLevel = selectors.get(selector);
```
nzakas, line 640: This can be undefined if someone is using an unknown selector. In that case, we just want to exit without reporting anything.

src/rules/require-baseline.js
```diff
@@ -625,6 +628,48 @@ export default {
 					});
 				}
 			},
+
+			Selector(node) {
```
nzakas, line 632: This matches all selectors, including those in `@supports`:

```css
@supports selector(:nth-child(1 of .foo)) {
  /* Styles */
}
```

When in `@supports`, we should not warn because they are checking for browser support.

Also related to `@supports`: If someone is checking if a selector is supported then we should not warn on any selectors used inside of that `@supports`. We already do this for property names and values, so we should do the same for selectors.

nzakas (changes requested):
Thanks for getting this started. The basics are here, there are a couple of edge cases that need addressing.

src/rules/require-baseline.js
rviscomi: Is this addressed by the check above?

```js
if (!selectors.has(selector)) {
  continue;
}
```

src/rules/require-baseline.js
rviscomi: Based on https://github.com/eslint/css/pull/61#issuecomment-2676252875 it doesn't seem like CSSTree parses selectors within `@supports` preludes. So if I understand correctly, while we should be able to tell whether the selector is used somewhere within a `@support` rule, we won't necessarily know if the selector was used in the `@support` condition itself.

For example, if an unrelated property is used in the condition, the non-Baseline feature could still be used in the declaration block:

```css
@supports (color: red) {
	h1:has(+ h2) {
		color: red;
	}
}
```

Is that an acceptable false negative? If so I can update the code and add a note in the docs.

rviscomi: Ah I just realized that the astexplorer.net tool I was using to check the AST is stuck on v2.0.1 and the selector support we need was added to v3.0.0. I'll take another look at this.

rviscomi: Updated

src/rules/require-baseline.js
nzakas: Ah yes, sorry about that.

src/rules/require-baseline.js
nzakas: Yup, I'd recommend using https://explorer.eslint.org, which uses the same parser version as this plugin.

nzakas:
This is looking really good. I think the last step here is to update the docs with some examples that use `@supports`.

rviscomi:
Great thanks, added another example to the docs.

nzakas (approved):
LGTM. Thanks so much for this. Great work.

Issue #13, opened 2024-11-20
New Rule: use-logical


NateRadebaugh:
### Rule details

Enforce the usage of [Logical Properties and Values](https://www.w3.org/TR/css-logical-1/) in CSS.  Physical dimensions and directions are described left to right and top to bottom, while their logical counterparts are described start to end and inline or block.

### What type of rule is this?

Warns about a potential problem

### Example code

https://github.com/csstools/stylelint-use-logical

For example, to add spacing before the start of a paragraph, we might use the physical padding-left property.

```css
p {
  padding-left: 2em;
}
```

Were the content Hebrew or Arabic — flowing right to left — then we might use alternating padding-left and padding-right properties.

```css
p:dir(ltr) {
  padding-left: 2em;
}

p:dir(rtl) {
  padding-right: 2em;
}
```

Selector weight aside, we can instead use the logical padding-inline-start property.

```css
p {
  padding-inline-start: 2em;
}
```

### Participation

- [x] I am willing to submit a pull request to implement this rule.

### Additional comments

I'm not sure the quirks of the new css parser and context, but my current stylelint configuration for angular projects includes:

```json
{
  "extends": "stylelint-config-standard-scss",
  "plugins": ["stylelint-use-logical"],
  "overrides": [
    {
      "files": ["*.html", "**/*.html"],
      "customSyntax": "postcss-html"
    },
    {
      "files": ["**/*.component.ts"],
      "customSyntax": "postcss-angular"
    }
  ],
  "rules": {
    "csstools/use-logical": [
      "always",
      {
        "except": [
          "min-width",
          "max-width",
          "width",
          "min-height",
          "height",
          "max-height",
          "top",
          "bottom",
          "margin-top",
          "margin-bottom",
          "padding-top",
          "padding-bottom",
          "border-top",
          "border-bottom",
          "border-top-color",
          "border-bottom-color"
        ]
      }
    ]
  }
}
```

jeddy3:
There's an alternative Stylelint plugin called [`stylelint-plugin-logical-css`](https://www.npmjs.com/package/stylelint-plugin-logical-css) that lists a fuller set of logical properties, [values](https://drafts.csswg.org/css-logical-1/#directional-keywords) and [units](https://drafts.csswg.org/css-values/#viewport-relative-lengths) in its README.

For example:

```css
h1 { 
  font-size: 8vi;
  margin-block: 2vb;
  text-align: start;
}
```

[...]

PR #63, opened 2025-02-22
feat: add prefer-logical-properties rule

azat-io:
#### Prerequisites checklist

Add new rule.

Added new rule for usage logical CSS properties instead of phisical.

#### Related Issues

#13

I don't think so.

### DIFF
--- a/docs/rules/prefer-logical-properties.md
+++ b/docs/rules/prefer-logical-properties.md
@@ -33,3 +33,11 @@ This rule accepts an option object with the following properties:
 
 - `allowProperties` (default: `[]`) - Specify an array of physical properties that are allowed to be used.
 - `allowUnits` (default: `[]`) - Specify an array of physical units that are allowed to be used.
+
+## When Not to Use It
+
+If you aren't concerned with the use of logical properties, then you can safely disable this rule.
+
+## Prior Art
+
+- [stylelint-use-logical](https://github.com/csstools/stylelint-use-logical)
--- a/docs/rules/prefer-logical-properties.md
+++ b/docs/rules/prefer-logical-properties.md
@@ -26,3 +26,10 @@ a {
 	margin-inline-start: 10px;
 }
 ```
+
+### Options
+
+This rule accepts an option object with the following properties:
+
+- `allowProperties` (default: `[]`) - Specify an array of physical properties that are allowed to be used.
+- `allowUnits` (default: `[]`) - Specify an array of physical units that are allowed to be used.
--- a/docs/rules/prefer-logical-properties.md
+++ b/docs/rules/prefer-logical-properties.md
@@ -1,0 +1,28 @@
+# prefer-logical-properties
+
+Prefer logical properties over physical properties.
+
+## Background
+
+Logical properties are a set of CSS properties that map to their physical counterparts. They are designed to make it easier to create styles that work in both left-to-right and right-to-left languages. Logical properties are useful for creating styles that are more flexible and easier to maintain.
+
+## Rule Details
+
+This rule checks for the use of physical properties and suggests using their logical counterparts instead.
+
+Examples of **incorrect** code for this rule:
+
+```css
+/* incorrect use of physical properties */
+a {
+	margin-left: 10px;
+}
+```
+
+Examples of **correct** code for this rule:
+
+```css
+a {
+	margin-inline-start: 10px;
+}
+```
--- a/src/rules/prefer-logical-properties.js
+++ b/src/rules/prefer-logical-properties.js
@@ -133,13 +133,13 @@ export default {
 			{
 				type: "object",
 				properties: {
-					allowedProperties: {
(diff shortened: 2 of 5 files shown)


nzakas:
Thanks for getting this started. There were some edge cases that were discussed in the original issue:

https://github.com/eslint/css/issues/13#issuecomment-2488917780
https://github.com/eslint/css/issues/13#issuecomment-2489258859

Can you describe how you're handling those?

nzakas:
I think we need a way for people to opt-out on a case-by-case basis. At a minimum, I think we'd need to be able to do this:

```js
{
    "prefer-logical-properties": ["error", {
        allowedProperties: ["padding-left"],   // don't warn on these
        allowedUnits: ["xyz"]  // don't warn on these
    }
}
```

src/rules/prefer-logical-properties.js
```diff
@@ -0,0 +1,175 @@
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const propertiesReplacements = {
```
nzakas, line 5: Please move all of these lookup tables into maps for better performance.

azat-io: Done ✅

azat-io:
I've added options `allowedUnits` and `allowedProperties` ✅.

src/rules/prefer-logical-properties.js
```diff
@@ -0,0 +1,232 @@
...
+]);
+
+//-----------------------------------------------------------------------------
+// Rule Definition
+//-----------------------------------------------------------------------------
+export default {
+	meta: {
+		type: /** @type {const} */ ("problem"),
+
+		fixable: "code",
```
nzakas, line 125: To fix the type error.
```suggestion
		fixable: /** @type {const} */ ("code"),
```

azat-io: Fixed ✅

azat-io:
Renamed the options and added their descriptions to the documentation.

docs/rules/prefer-logical-properties.md
nzakas, line 1: Just to finish up the docs, we typically include two additional sections:

1. When Not to Use It - gives guidance on when it's okay to not use the rule
2. Prior Art - links to other tools that do the same thing

Here's an example:
https://github.com/eslint/css/blob/main/docs/rules/no-invalid-properties.md

For prior art, we have:

* https://github.com/csstools/stylelint-use-logical
* https://www.npmjs.com/package/stylelint-plugin-logical-css

azat-io: Added ✅

nzakas:
This is looking good. I just realized that it will also error for `@supports` rules, such as:

```css
@supports (text-align: left) {
    /* rules */
}
```

I'm not sure if that's the correct behavior. What do you think?

jeddy3:
Looking good! Two questions...

Do we want the rule to also flag the [`width`](https://drafts.csswg.org/css-conditional-5/#width) and [`height`](https://drafts.csswg.org/css-conditional-5/#descdef-container-height) [size features](https://drafts.csswg.org/css-conditional-5/#size-container) (useful for alignment with the [container type](https://drafts.csswg.org/css-conditional-5/#container-type))?, i.e. allow this:

```css
main {
  container-type: inline-size;

  & h2 {
    @container (inline-size > 40em) {
      /* .. */
    }
  }
```

While disallowing this:


```css
main {
  container-type: inline-size;

  & h2 {
    @container (width > 40em) {
      /* .. */
    }
  }
```

And do we want to do anything special with `overflow-x` and `overflow-y` as their [logical counterparts are only supported by Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-inline#browser_compatibility)?

nzakas:
> And do we want to do anything special with `overflow-x` and `overflow-y` as their [logical counterparts are only supported by Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-inline#browser_compatibility)?

I think we can skip these for now. If/when other browsers implement the logical equivalent we can revisit.

> Do we want the rule to also flag the [`width`](https://drafts.csswg.org/css-conditional-5/#width) and [`height`](https://drafts.csswg.org/css-conditional-5/#descdef-container-height) [size features](https://drafts.csswg.org/css-conditional-5/#size-container) (useful for alignment with the [container type](https://drafts.csswg.org/css-conditional-5/#container-type))?,

These aren't properties, so I'm not sure it makes sense to bundle this check in this rule.

azat-io:
> This is looking good. I just realized that it will also error for @supports rules

You are right. We should not use linting in `@supports` directives. Stylelint's rules don't do that either.

I've updated the PR and added new tests.

image

nzakas (approved):
LGTM. Thanks for all of your work on this.

(The CI failure is unrelated to this change, so we're not going to worry about that.)

PR #81, opened 2025-03-11
feat: add support for baseline years

### LINKED ISSUE #78: Change Request: add support for Baseline years to require-baseline
### Environment

Node version: v20.0.0
npm version: v9.6.4
Local ESLint version: v9.21.0 (Currently used)
Global ESLint version: Not found
Operating System: darwin 23.6.0

### What problem do you want to solve?

In addition to Baseline newly and widely available, features can also be expressed in terms of the year they became newly available. For example, here is the list of features considered [Baseline 2023](https://webstatus.dev/?q=baseline_date%3A2023-01-01..2023-12-31). This can be useful for developers who prefer to set a fixed browser support target that they can periodically reevaluate, rather than a moving target like widely available. However, the require-baseline rule currently only supports two modes: newly or widely available.

### What do you think is the correct solution?

I'd like the require-baseline rule to also support Baseline years. So a user can say that they target features released in or before a given year, eg `available: 2021`, and the rule will not allow any features newer than that.

### Participation

- [x] I am willing to submit a pull request for this change.

### Additional comments

Related to https://github.com/eslint/css/discussions/67, because Baseline years are fixed targets, they don't have the same transience issues as tests targeting newly available features.

rviscomi:
#### Prerequisites checklist

Adds the ability to configure the require-baseline rule to target a Baseline year as an alternative to newly or widely available status

- added the integer type as an alternative to the `available` configuration property enum
- created a `BaselineAvailability` class to determine whether a feature is supported based on the config
- added tests that use the year config
- generated feature years in addition to their status ID
- regenerated baseline-data.js
- updated docs
- ran the formatter, which generated an unrelated change to README.md

#### Related Issues

fixes #78 

For the rule schema, would you prefer to configure the baseline year as a separate optional property, rather than shoehorning it in with `available`?


nzakas:
Thanks for putting this together. I'm not sure how useful this feature is, so let's discuss back on #78 before iterating on the code.

src/rules/require-baseline.js
```diff
@@ -339,6 +339,35 @@ class SupportsRules {
...
+	 */
+	constructor(availability) {
+		this.availability = availability;
+
+		if (Number.isInteger(availability)) {
+			this.baselineYear = availability;
+		} else {
+			this.baselineStatus =
+				availability === "widely" ? BASELINE_HIGH : BASELINE_LOW;
+		}
```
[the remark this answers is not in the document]
nzakas: Also, this seems like a good place to validate `availability`. For numbers, it seems like we should check between the lower bound of baseline availability and the current year - 1?

src/rules/require-baseline.js
```diff
@@ -339,6 +339,35 @@ class SupportsRules {
 	}
 }
 
+class BaselineAvailability {
```
nzakas, line 342: Please add some JSDoc here explaining the purpose of this class.

src/rules/require-baseline.js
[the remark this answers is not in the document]
rviscomi: Made `baselineYear` and `baselineStatus` private fields. Kept `availability` public since that one's actually accessed externally.

I've also added validation boundaries at 2000 and YYYY+1. I went with +1 instead of -1 because while I personally would never encourage it, the current Baseline year is technically valid. Let me know if you disagree. I've also added it to the schema object - not sure if you were thinking of something else like a custom error message emitted with `context.report`.

rviscomi:
Thanks @nzakas this is ready for another look

> Do we want to wait to merge https://github.com/eslint/css/pull/82 before merging this?

No preference. Merge conflicts either way 😁

tools/generate-baseline.js
```diff
@@ -33,6 +33,19 @@ const baselineIds = new Map([
 	[false, BASELINE_FALSE],
 ]);
 
+/**
+ * Maps the raw feature status object to a baseline status ID and year.
+ * @param {Object} status The raw feature status object.
+ * @returns {Object} An object containing the baseline status and year.
+ */
+function mapFeatureStatus(status) {
+	return {
```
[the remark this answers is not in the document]
rviscomi: Minimizing memory makes sense.

One thing we lose is the readability of the data file, as features would have opaque statuses like `8082` instead of `{status: 10, year: 2020}`. Looking through that file is actually how I spotted an issue with one of the property values in https://github.com/eslint/css/pull/82#discussion_r1991403309, so it seems like a useful quality.

Is there an approach that optimizes memory and readability? Maybe something like:

```js
const encodedStatus = `${status} (${year})`;
```

We could also omit the year when the status is 0, since it'll always be null. Either way, this introduces an additional parsing step, and I'm not sure how much memory it saves over the original implementation.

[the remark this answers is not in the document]
nzakas: Doing it as a single string would work, too. :+1:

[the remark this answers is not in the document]
rviscomi: Updated

src/rules/require-baseline.js
```diff
@@ -339,6 +339,21 @@ class SupportsRules {
...
+ * Decodes and extracts the status and year from an encoded feature status.
+ * @param {string} encodedStatus The encoded feature status and year.
+ * @returns {Object} An object containing the feature status and year.
+ */
+function decodeBaselineStatus(encodedStatus) {
+	const match = encodedStatus.match(
+		/^(?<status>\d+)(?: \((?<year>\d{4})\))?$/u,
+	);
+	const status = Number(match.groups.status);
+	const year = Number(match.groups.year);
```
nzakas, line 352: I think we can move these lines directly into the `isSupported()` method. That way, we can save the object created as the return value and we can also avoid running this same logic twice for each feature. What do you think?

tools/generate-baseline.js
```diff
@@ -33,17 +33,31 @@ const baselineIds = new Map([
 	[false, BASELINE_FALSE],
 ]);
 
+/**
+ * Encodes the baseline status and year fields into a single string.
+ * @param {{status: string, year: number}} baselineStatus The baseline fields to encode.
+ * @returns {string} The encoded baseline status and year.
+ */
+function encodeBaselineStatus({ status, year }) {
```
nzakas, line 41: Can we change `status` and `year` to regular (not destructured) parameters? There's no need to create an object just to call this function.

tools/generate-baseline.js
```diff
@@ -33,17 +33,31 @@ const baselineIds = new Map([
...
+	if (status === BASELINE_FALSE) {
+		// by definition, this status does not have a year, so omit it
+		return String(BASELINE_FALSE);
+	}
+
+	return `${status} (${year})`;
```
nzakas, line 47: I'm not sure how valuable the parentheses are here. Maybe for simpler parsing we could do this?
```suggestion
	return `${status}:${year}`;
```

Then we can just split on ":" and avoid regexs altogether.

nzakas (commented):
Almost there, just left a few comments around optimization.

src/rules/require-baseline.js
rviscomi: The logic should only run one time per feature, thanks to the early return.

Something like might be clearer:

```js
isSupported(feature) {
	const { status, year } = decodeBaselineStatus(feature);
	
	if (this.#baselineYear) {
		return year <= this.#baselineYear;
	}

	return status >= this.#baselineStatus;
}
```

Either way, I'm happy to merge this with `isSupported` if you still prefer.

rviscomi: Ok, updated

src/rules/require-baseline.js
```diff
@@ -357,7 +405,17 @@ export default {
...
-						enum: ["widely", "newly"],
+						anyOf: [
+							{
+								enum: ["widely", "newly"],
+							},
+							{
+								// baseline year
+								type: "integer",
+								minimum: 2000,
+								maximum: new Date().getFullYear() + 1,
```
nzakas, line 416: Double-checking: is this intentional? The max now would be 2026?

Or did you mean this?
```suggestion
								maximum: new Date().getFullYear() - 1,
```

rviscomi: Ah, yes this was intentional (see https://github.com/eslint/css/pull/81#discussion_r1992486853) but I just realized the maximum is inclusive. My reasoning was that the current year is technically a valid Baseline year, ie https://web.dev/baseline/2025.

Fixed this to use the current year, but let me know if you'd still prefer `YYYY - 1`.

nzakas: Sounds good to me. :+1:

nzakas (approved):
LGTM. Thanks for all of your work on this.

PR #125, opened 2025-05-06
feat: add rule types

thecalamiity:
#### Prerequisites checklist

This pull request introduces the `CSSRules` interface, which defines the TypeScript types for all the rules available in this package.

- Created `src/types/rules.d.ts`. It defines the `CSSRules` interface.
- Added `tools/update-rule-type-headers.js`. This script automates TSDoc header comments in the type definition file. It was adapted from ESLint's repo. Edits include rule fetching and ESM conversion.
- Added a `lint:rule-types` script to `package.json` to check if the rule types are up-to-date.
- Updated the CI workflow to include this check.
- Configured `rollup.config.js` to output the type definitions in both cjs and esm formats.
- Added a `pre-commit` hook to automatically update rule type headers when rule files are changed.

If the team approves these changes, I will implement similar type definitions and tooling for `@eslint/json` and `@eslint/markdown`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,18 +36,20 @@ In your `eslint.config.js` file, import `@eslint/css` and include the recommende
 
 ```js
 // eslint.config.js
+import { defineConfig } from "eslint/config";
 import css from "@eslint/css";
 
-export default [
+export default defineConfig([
 	// lint CSS files
 	{
 		files: ["**/*.css"],
 		language: "css/css",
-		...css.configs.recommended,
+		plugins: { css },
+		extends: ["css/recommended"],
 	},
 
 	// your other configs here
-];
+]);
 ```
 
 ### Rules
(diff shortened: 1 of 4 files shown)


nzakas:
Thanks for exploring this. Before moving forward, I think we should step back and put together an RFC about how we want to expose type information from language plugins. There are a lot of moving pieces here, and I think we may want to move some of the dependent types into `@eslint/core` rather than using them from `eslint`.

(In general, when adding something new to a package, it's best to open an issue to explain your idea before starting a PR.)

Suggested next steps:

1. Open an issue in the `eslint` repo for "Establish guidelines for language plugin rule types" (or however you want to phrase it). Include your idea of how to do this.
2. Once we've had some discussion, then it would be good to create an RFC so we can be sure to get feedback from the community before deciding on a solution.

How does that sound?

PR #137, opened 2025-05-15
fix: improve use-baseline types regex

### LINKED ISSUE #135: Bug: `use-baseline` reports `linear-gradient` and `radial-gradient` as not being widely available
### Environment

ESLint version: 9.26.0
@eslint/css version: 0.8.0
Node version: 22.15.1
npm version: 11.4.0
Operating System: Ubuntu & Windows 11 Pro

stylesheet

### What did you do?

<details>

```js
import { defineConfig } from "eslint/config";

export default defineConfig([
	// lint CSS files
	{
		files: ["**/*.css"],
		language: "css/css",
		plugins: { css },
		extends: ["css/recommended"],
	},

	// your other configs here
]);

```
</details>

```js
body {
  background-size: 100vi 100vb;
  background-image:
    linear-gradient(180deg,
      var(--color-bg-0) 0%,
      var(--color-bg-1) 15%,
      var(--color-bg-2) 50%) /* should be OK, but errors */;
    radial-gradient(50% 50% at 50% 50%,
      rgba(255, 255, 255, 0.75) 0%,
      rgba(255, 255, 255, 0) 100%) /* should be OK, but errors */,
}

```

I expected the above code to not produce any error.

## Reasoning

[`linear-gradient`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient) and [`radial-gradient`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient) are considered widely-available by `mdn`, with few exceptions noted [here](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient#browser_compatibility) and [here](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient#browser_compatibility).

### What actually happened?

Received the following errors (with screenshot):

> Type 'linear-gradient' is not a widely available baseline feature.eslint[css/use-baseline](https://github.com/eslint/css/blob/main/docs/rules/use-baseline.md)

> Type 'radial-gradient' is not a widely available baseline feature.eslint[css/use-baseline](https://github.com/eslint/css/blob/main/docs/rules/use-baseline.md)

Image

Please use the exact config and exact code above for minimum repo

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Additional comments

_No response_

rviscomi:
#### Prerequisites checklist

Fixes a bug in the `types` regex that was matching too many BCD keys

Simplified the regex so that it always matches the last field of the key. Anything that's not a function will get filtered out later in the comparison with `mdnData.css.functions`.

#### Related Issues

fixes #135

If you could, spot check some of the functions' Baseline status/year values against the corresponding MDN docs. For example, `linear-gradient` has a value of `10:2015` (widely available, Baseline since 2015), which we also see on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient):

image


nzakas (approved):
LGTM. Thanks for digging into this.

I checked each of the changes in the diff and verified the data is correct.

jimmy-zhening-luo:
Thank you very much!

PR #168, opened 2025-06-08
fix: tailwind-syntax should allow plugin atrule

tisonkun:
For example:

```css
@import "tailwindcss";
@plugin '@tailwindcss/typography';

```


tisonkun:
@lumirlumir The failure seems to be unrelated to this patch.

PR #169, opened 2025-06-09
feat: add no-invalid-named-grid-areas rule

### LINKED ISSUE #158: New Rule: no-invalid-named-grid-areas
### Rule details

Disallow invalid named grid areas.

### Example code

```css
a { grid-template-areas: "a a a"
                         "b b b b"; }
```

### Prior Art

[named-grid-areas-no-invalid](https://stylelint.io/user-guide/rules/named-grid-areas-no-invalid/) from stylelint

### Participation

### Additional comments

For a named grid area to be valid, all strings must define:

- the same number of cell tokens
- at least one cell token

And all named grid areas that spans multiple grid cells must form a single filled-in rectangle.

xbinaryx:
#### Prerequisites checklist

This PR adds a new rule `no-invalid-named-grid-areas` to enforce valid named grid area definitions in CSS Grid templates. The rule helps prevent common mistakes when using grid-template-areas and related properties by checking that named grid areas are properly defined according to the CSS Grid specification.

Added the  `no-invalid-named-grid-areas` rule, along with documentation and tests.

#### Related Issues

Fixes #158

### DIFF
--- a/src/rules/no-invalid-named-grid-areas.js
+++ b/src/rules/no-invalid-named-grid-areas.js
@@ -29,6 +29,7 @@ const nullCellToken = /^\.+$/u;
  */
 function findNonRectangularAreas(grid) {
 	const errors = [];
+	const reported = new Set();
 	const names = [...new Set(grid.flat())].filter(
 		name => !nullCellToken.test(name),
 	);
@@ -59,8 +60,11 @@ function findNonRectangularAreas(grid) {
 					row1.length !== row2.length ||
 					!row1.every((val, idx) => val === row2[idx])
 				) {
-					const firstRow = grid.findIndex(r => r.includes(name));
-					errors.push({ name, row: firstRow });
+					const key = `${name}|${j}`;
+					if (!reported.has(key)) {
+						errors.push({ name, row: j });
+						reported.add(key);
+					}
 				}
 			}
 		}
--- a/src/rules/no-invalid-named-grid-areas.js
+++ b/src/rules/no-invalid-named-grid-areas.js
@@ -1,0 +1,174 @@
+/**
+ * @fileoverview Rule to prevent invalid named grid areas in CSS grid templates.
+ * @author xbinaryx
+ */
+
+//-----------------------------------------------------------------------------
+// Type Definitions
+//-----------------------------------------------------------------------------
+
+/**
+ * @import { CSSRuleDefinition } from "../types.js"
+ * @typedef {"emptyGridArea" | "unevenGridArea" | "nonRectangularGridArea"} NoInvalidNamedGridAreasMessageIds
+ * @typedef {CSSRuleDefinition<{ RuleOptions: [], MessageIds: NoInvalidNamedGridAreasMessageIds }>} NoInvalidNamedGridAreasRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+/**
+ * Regular expression to match null cell tokens (sequences of one or more dots)
+ */
+const nullCellToken = /^\.+$/u;
+
+/**
+ * Finds non-rectangular grid areas in a 2D grid
+ * @param {string[][]} grid 2D array representing the grid areas
+ * @returns {Array<{name: string, row: number}>} Array of errors found
+ */
+function findNonRectangularAreas(grid) {
+	const errors = [];
+	const names = [...new Set(grid.flat())].filter(
+		name => !nullCellToken.test(name),
+	);
+
+	for (const name of names) {
+		const indicesByRow = grid.map(row => {
+			const indices = [];
+			let idx = row.indexOf(name);
+
+			while (idx !== -1) {
+				indices.push(idx);
+				idx = row.indexOf(name, idx + 1);
+			}
+
+			return indices;
+		});
+
+		for (let i = 0; i < indicesByRow.length; i++) {
+			for (let j = i + 1; j < indicesByRow.length; j++) {
+				const row1 = indicesByRow[i];
+				const row2 = indicesByRow[j];
+
+				if (row1.length === 0 || row2.length === 0) {
+					continue;
+				}
+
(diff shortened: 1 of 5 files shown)


nzakas (dismissed):
LGTM. Thanks! Would like another review before merging.

src/rules/no-invalid-named-grid-areas.js
```diff
@@ -0,0 +1,174 @@
...
+						}
+
+						const row = trimmedValue.split(" ").filter(Boolean);
+						grid.push(row);
+
+						if (firstRowLen === null) {
+							firstRowLen = row.length;
+						} else if (row.length !== firstRowLen) {
+							unevenNodes.push(stringNode);
+						}
```
snitin315, line 140: Right now we don't report the `"a b"` line, but I think we should:

```css
/* eslint css/no-invalid-named-grid-areas: "error" */

.grid {
	grid-template-areas:
  "a b" // No error, but shouldn't we report this as well since this is also uneven and we don't know if the intention is to have 2,3, or 4 cells
  "a b c" //  Error: Grid area strings must have the same number of cell tokens 
  "a b c d"; //  Error: Grid area strings must have the same number of cell tokens 
}
```

snitin315: Because the grid is inconsistent as a whole.

src/rules/no-invalid-named-grid-areas.js
```diff
@@ -0,0 +1,174 @@
...
+					nonRectErrors.forEach(({ name, row }) => {
+						const stringNode = stringNodes[row];
+						context.report({
+							node: stringNode,
+							messageId: "nonRectangularGridArea",
+							data: {
+								name,
+							},
+						});
+					});
```
snitin315, line 173: Here we report an error on line 5, but I think we should report line 6:

Screenshot 2025-06-21 at 9 48 24 PM

snitin315: Here error in both rows but we report only one:

Screenshot 2025-06-21 at 9 50 25 PM

src/rules/no-invalid-named-grid-areas.js
```diff
@@ -0,0 +1,174 @@
...
+
+					for (const stringNode of stringNodes) {
+						const trimmedValue = stringNode.value.trim();
+
+						if (trimmedValue === "") {
+							emptyNodes.push(stringNode);
+							continue;
+						}
+
+						const row = trimmedValue.split(" ").filter(Boolean);
```
snitin315, line 133: Apparently, the escaping whitespace is valid syntax:

Screenshot 2025-06-21 at 9 58 33 PM

```css
grid-template-areas:
  "a \ a a" // Throws error but is valid
  "b c c"
  "b c c";
```

tests/rules/no-invalid-named-grid-areas.test.js
```diff
@@ -0,0 +1,256 @@
...
+
+const ruleTester = new RuleTester({
+	plugins: {
+		css,
+	},
+	language: "css/css",
+});
+
+ruleTester.run("no-invalid-named-grid-areas", rule, {
+	valid: [
```
snitin315, line 27: Let's add some valid test cases with whitespace also:

```css
grid-template-areas:
  "  a b c "
  "a b c";
  
grid-template-areas:
  "header     header"
  "main       sidebar";
```

src/rules/no-invalid-named-grid-areas.js
nzakas: For most linting rules we only flag the first pattern that is unexpected. In this example, the second row is unexpected because of the first row setting the expectation that a row has two cells. So I think the current behavior is inline with how we generally make lint rules work.

src/rules/no-invalid-named-grid-areas.js
nzakas: As with my other comment, I think it's correct to only error on the second row here.

src/rules/no-invalid-named-grid-areas.js
xbinaryx: I couldn't reproduce this. Can you double-check?

src/rules/no-invalid-named-grid-areas.js
snitin315: Agreed, but here we are reporting the first row instead of 2nd in both cases.

src/rules/no-invalid-named-grid-areas.js
snitin315: Makes sense 👍🏻

src/rules/no-invalid-named-grid-areas.js
snitin315: Hmm, can't reproduce anymore.

src/rules/no-invalid-named-grid-areas.js
xbinaryx: @snitin315 Actually, it does report the second row in both cases. Could you please double-check?

snitin315: @xbinaryx Can you please add a corresponding test cse?

tests/rules/no-invalid-named-grid-areas.test.js
```diff
@@ -0,0 +1,288 @@
...
+				{
+					messageId: "nonRectangularGridArea",
+					data: { name: "a" },
+					line: 1,
+					column: 32,
+					endLine: 1,
+					endColumn: 37,
+				},
+			],
+		},
```
snitin315: Can we format it like https://github.com/eslint/css/pull/169#discussion_r2160075776 to clearly see which row is actually being reported

```css
grid-template-areas: 
     "a a"
      "a .";
```

It should report the 2nd row.

nzakas (approved):
LGTM. Would like @snitin315 to verify before merging.

snitin315 (approved):
LGTM, thanks!

PR #171, opened 2025-06-11
feat: add no-invalid-at-rule-placement rule

### LINKED ISSUE #151: New Rule: no-invalid-import-position
### Rule details

Disallow invalid placement of @import rules.

### Example code

```css
a {}
@import 'foo.css';
```

### Participation

### Additional comments

`@import` rules must precede all other valid at-rules and style rules in a stylesheet (ignoring `@charset` and `@layer`), or else the `@import` rule is invalid.

Prior Art:
https://stylelint.io/user-guide/rules/no-invalid-position-at-import-rule/

### LINKED ISSUE #163: New Rule: no-invalid-charset-placement
### Rule details

Disallow invalid placement of @charset rule.

### Example code

```css
a {}
@charset "utf-8";
```

### Prior Art

```markdown
I couldn't find a Stylelint equivalent, but VS Code's built-in linter reports this.
```

### Participation

### Additional comments

`@charset` can only be placed at the very beginning of a stylesheet. No other characters, except the Unicode byte-order mark, are allowed before it.

thecalamiity:
#### Prerequisites checklist

This pull request introduces a new rule no-invalid-at-rule-placement that enforces correct placement of at-rules within stylesheets.

Implemented the no-invalid-at-rule-placement rule, along with documentation and tests.

#### Related Issues

Fixes #163
Fixes #151


docs/rules/no-invalid-at-rule-placement.md
```diff
@@ -0,0 +1,95 @@
+# no-invalid-at-rule-placement
+
+Disallow invalid placement of at-rules.
+
+## Background
+
+At-rules are CSS statements that instruct CSS how to behave. Some at-rules have strict placement requirements that must be followed for the stylesheet to work correctly. For example:
```
jeddy3, line 7: `@charset` is [no longer considered an at-rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset) even though it syntactically looks like one. There may be a more suitable name and wording in the documentation that better reflect this.

jeddy3 (commented):
Nice and useful rule that plugs a gap, as CSSTree only detected misplaced declarations, e.g.:

```css
--foo: bar; /* CSSTree parse error */
a { }
```

docs/rules/no-invalid-at-rule-placement.md
nzakas: If it's not an at-rule, then what is it? The spec is ridiculously worded:
https://drafts.csswg.org/css-syntax/#charset-rule

thecalamiity: According to MDN, it's not technically an at-rule but a specific byte sequence. However, I noticed it's referred to as an at-rule in no-invalid-at-rules as well. Should we add a clarifying note about this, or leave it as-is for now?

nzakas: Let's leave this as-is for now. We can always circle back and do doc updates later.

docs/rules/no-invalid-at-rule-placement.md
```diff
@@ -0,0 +1,95 @@
...
+@import "bar.css";
+```
+
+Examples of **correct** code:
+
+```css
+/* eslint css/no-invalid-at-rule-placement: "error" */
+
+/* @charset at the beginning */
+@charset "utf-8";
```
snitin315: Let's move the comment position because event comments can't be added before `@charset`.

```suggestion
@charset "utf-8"; /* @charset at the beginning */
```

src/rules/no-invalid-at-rule-placement.js
```diff
@@ -0,0 +1,85 @@
...
+						context.report({
+							node,
+							messageId: "invalidCharsetPlacement",
+						});
+					}
+					return;
+				}
+
+				if (name === "layer") {
+					if (node.block) {
```
snitin315, line 61: Why is block reuired?

tests/rules/no-invalid-at-rule-placement.test.js
```diff
@@ -0,0 +1,267 @@
...
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+	plugins: {
+		css,
+	},
+	language: "css/css",
+});
+
+ruleTester.run("no-invalid-at-rule-placement", rule, {
```
snitin315, line 26: Can we add test case like following:

```css
@import "a.css";
@layer a;
@import "b.css";
```

docs/rules/no-invalid-at-rule-placement.md
JoshuaKGoldberg: Interestingly, I think this is arguably valid even without proper uses of `@charset` being at-rules. From the docs:

> If a `@charset` is not recognized as the charset declaration, it is parsed as a normal at-rule. 

...so an improper `@charset` is roughly speaking, an invalid at-rule placement. Sort of.

docs/rules/no-invalid-at-rule-placement.md
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+## When Not to Use It
+
+You can disable this rule if your stylesheets don't use `@charset` or `@import` rules, or if you're not concerned about the impact of incorrect placement on encoding and style loading.
+
+## Prior Art
+
+- [`no-invalid-position-at-import-rule`](https://stylelint.io/user-guide/rules/no-invalid-position-at-import-rule/)
```
JoshuaKGoldberg, line 94: [Docs] I would suggest adding further reading links for each of the linted at-rule/declaration/statement types. Even without the confusing points of `@charset`s.

```suggestion
- [`no-invalid-position-at-import-rule`](https://stylelint.io/user-guide/rules/no-invalid-position-at-import-rule/)

## Further Reading

- [@charset - CSS | MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset): `@charset` is not technically an at-rule when used properly. However, improper uses are parsed as at-rules.
- [@import - CSS | MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset)
```

src/rules/no-invalid-at-rule-placement.js
```diff
@@ -0,0 +1,85 @@
...
+
+		docs: {
+			description: "Disallow invalid placement of at-rules",
+			recommended: true,
+			url: "https://github.com/eslint/css/blob/main/docs/rules/no-invalid-at-rule-placement.md",
+		},
+
+		messages: {
+			invalidCharsetPlacement:
+				"@charset must be placed at the very beginning of the stylesheet, before any other rules, comments, or whitespace.",
```
JoshuaKGoldberg: [Docs] Per the thread about `@charset`, it's not _technically_ a rule. Which means we can get this nice small simplification of the report message:

```suggestion
				"@charset must be placed at the very beginning of the stylesheet, before any rules, comments, or whitespace.",
```

docs/rules/no-invalid-at-rule-placement.md
```diff
@@ -0,0 +1,95 @@
+# no-invalid-at-rule-placement
+
+Disallow invalid placement of at-rules.
+
+## Background
+
+At-rules are CSS statements that instruct CSS how to behave. Some at-rules have strict placement requirements that must be followed for the stylesheet to work correctly. For example:
+
+- The `@charset` rule must be placed at the very beginning of a stylesheet, before any other rules, comments, or whitespace.
+- The `@import` rule must be placed at the beginning of a stylesheet, before any other at-rules (except `@charset` and `@layer` statements) and style rules.
```
JoshuaKGoldberg, line 10: https://developer.mozilla.org/en-US/docs/Web/CSS/@namespace#description - `@namespace` also needs to be validated. Should we include that in this initial version of the rule as well?

snitin315: I think yes, we should add support for `@namespace` too, since this is a general rule, and the effort should not be high.

snitin315: @thecalamiity Can we include `@namespace`?

src/rules/no-invalid-at-rule-placement.js
thecalamiity: `@import` can be preceded by `@layer` statements but not `@layer` blocks.

docs/rules/no-invalid-at-rule-placement.md
thecalamiity: Agreed—supporting `@namespace` makes sense, but since this PR is ready, let’s address it in a separate change. Its logic would need to enforce placement after `@import` but before other rules, which introduces non-trivial complexity.

TKDev7: @thecalamiity I’d be happy to draft a PR for the `@namespace` validation

thecalamiity: Sounds good.

snitin315: Since we're already aware of the `@namespace` case and this rule is intended to cover all at-rules, I believe it makes sense to implement it in the same PR. 

That said, I'd love to hear the team's input. @eslint/eslint-team  do you think we should include it in this PR or address it in a follow-up?

nzakas: Since this PR is ready and there's already a draft PR to add `@namespace`, I'd say we move forward with this PR as-is.

snitin315: Cool, let's merge this then.

@TKDev7 Thanks for the draft. For future contributions, please create an issue or wait for team alignment before starting work on significant changes. This ensures everyone is on the same page and the effort aligns with the team's direction.

snitin315:
@thecalamiity There are few merge conflicts, can you please resolve?

nzakas:
@snitin315 this is ready for merge once you've verified everything.

PR #174, opened 2025-06-12
feat: add `font-family-fallbacks` rule

### LINKED ISSUE #153: New Rule: `require-fallback-fonts`
### Rule details

Rule will enforce the use of fallback font to maintain consistency and readability across different devices and browsers.

### Example code

```css
body {
  font-family: "Open Sans";
}
```

### Participation

### Additional comments

_No response_

Tanujkanti4441:
#### Prerequisites checklist

`Previous description` : This is a draft PR to strategize the variable support for the rule `font-family-fallbacks`.

`Latest description`: Add a new rule called `font-family-fallbacks` to enforce the use of fallback fonts and a generic font last.

`Previous description`: Added a rule called `font-family-fallbacks`. Right now, this PR just shows the strategy to enable variable support for this rule. So rule only contains the code to validate the use of `sans-serif` font in the variable (`--my-font`) used in `font-family`, and added two tests.

`Latest description`: Added a rule that will report following cases:

```css
/* eslint css/font-family-fallbacks: "error" */

:root {
	--my-font: "Open Sans";
}

a {
	font-family: Arial; /* error here, as only one font is specified. */
}

b {
	font-family: Verdana, Arial, Helvetica; /* error here, as no generic font is specified in last. */
}

.foo {
	font-family: var(--my-font); /* error here as only one font is specified in variable --my-font. */
}

a {
	font: 16px Arial; /* error here, as only one font is specified. */
}

b {
	font: italic bold 2rem var(--my-font); /* error here as only one font is specified in variable --my-font. */
}
```

#### Related Issues

fixes #153

### DIFF
--- a/src/rules/font-family-fallbacks.js
+++ b/src/rules/font-family-fallbacks.js
@@ -64,14 +64,14 @@ function reportFontWithoutFallbacksInFontProperty(
 
 		if (!containsGenericFont) {
 			context.report({
-				loc: node.value.loc,
+				loc: node.loc,
 				messageId: "useFallbackFonts",
 			});
 		}
 	} else {
 		if (!genericFonts.has(valueList.at(-1))) {
 			context.report({
-				loc: node.value.loc,
+				loc: node.loc,
 				messageId: "useGenericFont",
 			});
 		}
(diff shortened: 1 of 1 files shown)


docs/rules/font-family-fallbacks.md
```diff
@@ -0,0 +1,127 @@
...
+
+Enforce the use of fallback fonts and a generic font last.
+
+## Background
+
+The `font-family` property in CSS is used to specify which fonts should be used to display text. One can specify one or a list of fonts in `font-family`.
+
+The browser checks if the first font is available to the user's system, if the first font is not available, it tries the next font in the list. This process continues until a font is found. If none of the specified fonts are available, the browser uses the generic font family (like `serif`, `sans-serif`, etc.) at the end of the list.
+
+It is often suggested to use a fallback font and a generic font last in the `font-family` list to ensure that text is always displayed in readable and consistent way, even if the preferred fonts are not available on the user's system or if the browser doesn't support the preferred font.
```
nzakas:
```suggestion
It is a best practice to use a fallback font and a generic font last in the `font-family` list to ensure that text is always displayed in readable and consistent way, even if the preferred fonts are not available on the user's system or if the browser doesn't support the preferred font.
```

src/index.js
```diff
@@ -9,8 +9,18 @@
 
 import { CSSLanguage } from "./languages/css-language.js";
 import { CSSSourceCode } from "./languages/css-source-code.js";
-import recommendedRules from "./build/recommended-config.js";
-import rules from "./build/rules.js";
+import fontFamilyFallbacks from "./rules/font-family-fallbacks.js";
```
nzakas: It looks like this is reverting the recent work that we did to build up the rules list. Can you rebase and double-check?

src/rules/font-family-fallbacks.js
```diff
@@ -0,0 +1,484 @@
...
+			useGenericFont: "Use a generic font last.",
+		},
+	},
+
+	create(context) {
+		const sourceCode = context.sourceCode;
+		const variableMap = new Map();
+
+		return {
+			Rule(node) {
```
nzakas: This logic to find variables looks more complicated than necessary and potentially buggy. Please see how it's done in `no-invalid-properties`.

src/rules/font-family-fallbacks.js
```diff
@@ -0,0 +1,484 @@
...
+							const variableValue =
+								child.value.type === "Raw" && child.value.value;
+
+							variableMap.set(variableName, variableValue);
+						}
+					});
+				}
+			},
+
+			Declaration(node) {
```
nzakas: This function does a lot of traversing, which we try to avoid inside of rules. I think you can simplify this by using an ESQuery selector, like this:

```js
[Rule > Block > Declaration[property="font-family"] > Value](node) {
```

src/rules/font-family-fallbacks.js
```diff
@@ -0,0 +1,484 @@
...
+										lastNode.children[0].name;
+									const lastVariable =
+										variableMap.get(variableName);
+
+									if (!lastVariable) {
+										return;
+									}
+								}
+
+								node.value.children.forEach(child => {
```
nzakas: See if we can avoid multiple lookups like this (`a.b.c.d`) as the perf hit for continually looking up properties can add up over time.

src/rules/font-family-fallbacks.js
```diff
@@ -0,0 +1,484 @@
...
+					if (node.value.type === "Value") {
+						if (node.value.children.length === 1) {
+							// If it font is set to system font, we don't need to check for fallbacks
+							if (node.value.children[0].type === "Identifier") {
+								return;
+							}
+
+							// If the value is a variable function, we need to check the variable value
+							if (
+								node.value.children[0].type === "Function" &&
```
nzakas: Similar here, `node.value.children[0]` is used in a lot of places, so let's store that in a variable to avoid unnecessary lookups.

Tanujkanti4441: Done!

PR #178, opened 2025-06-19
feat: no-invalid-properties allowUnknownVariables option

### LINKED ISSUE #175: Rule Change: `no-invalid-properties` option to only validate property names with unknown variables

no-invalid-properties

Generate fewer warnings

A new option

### Example code

```css
.one {
	ccolor: var(--my-color);
}

.two {
	color: var(--my-color);
}
```

1. Can't validate with unknown variable '--my-color'
2. Can't validate with unknown variable '--my-color'

1. Unknown property 'ccolor' found
2. (no violation)

### Participation

### Additional comments

The changes made in 0.9.0 requires us to completely disable `no-invalid-properties` because our design system is based on a huge number of CSS variables that are defined in a single file. The guidance in the docs is not practical:

> If the custom property is defined in another file, it's recommended to create a dummy rule just to ensure proper validation.

This would be a huge lift and would non-trivially increase our CSS size if every variable _usage_ needed a `.dummy { --my-color: blue; }` declaration.

An option to ignore unknown variables and fall back to property name validation would be a great compromise for us.

jgoz:
#### Prerequisites checklist

Adds an option to `no-invalid-properties` called `allowUnknownVariables`. When true, variable references that can't be traced to a custom property definition are ignored.

Added an options schema for `no-invalid-properties` describing `allowUnknownVariables`. When the option is `true`, do the following:
- Ignore variable references that were not found up to this point (skip reporting `unknownVar`)
- Only report `unknownProperty` for errors of type `SyntaxReferenceError`, which indicate that the property is unknown

Also updated documentation for the rule to reflect this option as best I could.

#### Related Issues

Fixes #175


src/rules/no-invalid-properties.js
```diff
@@ -95,6 +108,9 @@ export default {
 		 */
 		const replacements = [];
 
+		const allowUnknownVariables =
+			context.options?.[0]?.allowUnknownVariables ?? false;
```
nzakas: If you use `defaultOptions` ([example](https://github.com/eslint/css/blob/b720cf89b611eca7b5778d11bdc4c872813209c3/src/rules/prefer-logical-properties.js#L171-L176)), then you can do this:

```suggestion
			context.options.[0].allowUnknownVariables;
```

nzakas (commented):
This is looking really good. Just one note to clean things up a bit.

src/rules/no-invalid-properties.js
jgoz: Ah, thank you. Updated, but used destructuring syntax to match the style in `relative-font-units`

snitin315:
Another case:

```css
/* eslint css/no-invalid-properties: ["error",  { allowUnknownVariables: false }] */

:root {
	--my-color: red;
}

.two {
	/* pink if --my-var and --my-background are not defined */
	color: var(--my-color, var(--my-background, pink));  //  error  Unknown property 'color' found  css/no-invalid-properties
}
```

jgoz:
> Thanks for the PR. I found one issue while [defining fallbacks in var](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties#defining_fallbacks_in_the_var_function)

Good catch, but I think this issue exists in main. Maybe it should be treated as a separate issue?

docs/rules/no-invalid-properties.md
```diff
@@ -44,9 +44,25 @@ body {
 }
 ```
 
-### Limitations
+## Options
 
-When a variable is used in a property value, such as `var(--my-color)`, the rule can only properly be validated if the parser has already encountered the `--my-color` custom property. For example, this will validate correctly:
+This rule accepts an option which is an object with the following property:
+
+- `allowUnknownVariables` (default: `false`) - Ignore variables that cannot be traced to custom properties in the current file
```
Tanujkanti4441:
```suggestion
- `allowUnknownVariables` (default: `false`) - Ignore variables that cannot be traced to custom properties in the current file.
```

docs/rules/no-invalid-properties.md
```diff
@@ -44,9 +44,25 @@ body {
...
+
+Examples of **incorrect** code with `{ allowUnknownVariables: false }` (the default):
+
+```css
+a {
+	color: var(--my-color);
+}
+```
+
+This code uses `var(--my-color)` before `--my-color` was defined, or where it is defined in another CSS file. Therefore, `color: var(--my-color)` cannot be property validated.
```
Tanujkanti4441:
```suggestion
This code uses `var(--my-color)` before `--my-color` is defined, or whether it is defined in another CSS file. Therefore, `color: var(--my-color)` cannot be properly validated.
```

docs/rules/no-invalid-properties.md
```diff
@@ -44,9 +44,25 @@ body {
...
 
-### Limitations
+## Options
 
-When a variable is used in a property value, such as `var(--my-color)`, the rule can only properly be validated if the parser has already encountered the `--my-color` custom property. For example, this will validate correctly:
+This rule accepts an option which is an object with the following property:
+
+- `allowUnknownVariables` (default: `false`) - Ignore variables that cannot be traced to custom properties in the current file
+
+When a variable is used in a property value, such as `var(--my-color)`, the rule can only properly be validated if the parser has already encountered the `--my-color` custom property. With `{ allowUnknownVariables: false }`, unknown variables will result in a lint error. With `{ allowUnknownVariables: true }`, the property value will be ignored and only the property name will be validated.
```
Tanujkanti4441:
```suggestion
When a variable is used in a property value, such as `var(--my-color)`, the rule can only properly be validated if the parser has already encountered the `--my-color` custom property. With `{ allowUnknownVariables: false }`, unknown variables will result in a linting error. With `{ allowUnknownVariables: true }`, the property value will be ignored and only the property name will be validated.
```

docs/rules/no-invalid-properties.md
```diff
@@ -44,9 +44,25 @@ body {
...
+
+Examples of **incorrect** code with `{ allowUnknownVariables: false }` (the default):
+
+```css
+a {
+	color: var(--my-color);
+}
```
Tanujkanti4441, line 62:
```suggestion
/* eslint css/no-invalid-properties: ["error", { allowUnknownVariables: false }] */

a {
	color: var(--my-color);
}
```

Can we add eslint config comments in `incorrect` and `correct` examples?

docs/rules/no-invalid-properties.md
```diff
@@ -44,9 +44,25 @@ body {
...
+```
+
+This code uses `var(--my-color)` before `--my-color` was defined, or where it is defined in another CSS file. Therefore, `color: var(--my-color)` cannot be property validated.
+
+Examples of **correct** code with `{ allowUnknownVariables: false }`:
```
Tanujkanti4441, line 67: `/* eslint css/no-invalid-properties: ["error", { allowUnknownVariables: false }] */`

in this example also.

docs/rules/no-invalid-properties.md
```diff
@@ -60,7 +76,25 @@ a {
 
 This code defines `--my-color` before it is used and therefore the rule can validate the `color` property. If `--my-color` was not defined before `var(--my-color)` was used, it results in a lint error because the validation cannot be completed.
 
-If the custom property is defined in another file, it's recommended to create a dummy rule just to ensure proper validation.
+Examples of **incorrect** code with `{ allowUnknownVariables: true }`:
+
+```css
+a {
+	ccolorr: var(--my-color);
+}
```
Tanujkanti4441, line 90:
```suggestion
/* eslint css/no-invalid-properties: ["error", { allowUnknownVariables: true }] */

a {
	ccolorr: var(--my-color);
}
```

docs/rules/no-invalid-properties.md
```diff
@@ -60,7 +76,25 @@ a {
...
+```
+
+This code uses an unknown property `ccolorr`, which results in a validation error. The unknown reference to `var(--my-color)` is ignored.
+
+Examples of **correct** code with `{ allowUnknownVariables: true }`:
+
+```css
+a {
+	color: var(--my-color);
+}
```
Tanujkanti4441, line 102:
```suggestion
/* eslint css/no-invalid-properties: ["error", { allowUnknownVariables: true }] */

a {
	color: var(--my-color);
}
```

Tanujkanti4441 (changes requested):

snitin315:
>Good catch, but I think this issue exists in main. Maybe it should be treated as a separate issue?

I see. Yes, let's tackle it separately.

docs/rules/no-invalid-properties.md
jgoz: Done

lumirlumir:
Hi everyone, The CI failure is due to the newly released Prettier v3.6.0.

I’ve opened #182 to fix the issue. Once it’s merged into main, the error should be resolved.

lumirlumir:
#182 has now been merged, so the CI error should be resolved once we merge the main branch. 👍

PR #190, opened 2025-07-06
build: automate generation of recommended rules config

xbinaryx:
#### Prerequisites checklist

Automate generation of the recommended rules config and rules export file.

- Added tools/build-rules.js script
- Script outputs src/build/recommended-config.js and src/build/rules.js based on rules in src/rules/


package.json
```diff
@@ -60,16 +60,17 @@
...
     "build:cts": "node -e \"fs.copyFileSync('dist/esm/index.d.ts', 'dist/cjs/index.d.cts')\" && node tools/update-cts.js dist/cjs/types.cts dist/cjs/index.d.cts",
-    "build": "rollup -c && npm run build:dedupe-types && tsc -p tsconfig.esm.json && npm run build:cts",
+    "build:rules": "node tools/build-rules.js",
+    "build": "npm run build:rules && rollup -c && npm run build:dedupe-types && tsc -p tsconfig.esm.json && npm run build:cts",
     "build:readme": "node tools/update-readme.js",
     "build:update-rules-docs": "node tools/update-rules-docs.js",
     "build:baseline": "node tools/generate-baseline.js",
-    "test:jsr": "npx jsr@latest publish --dry-run",
-    "pretest": "npm run build",
+    "prepare": "npm run build",
```
nzakas, line 68: This was `pretest` previously. Intentionally changed?

xbinaryx: Yes. We need to generate the `src/build` folder after `npm install`, and running the build once post-install is sufficient. There’s no need to rebuild before every `npm test`.

nzakas: The reason we had it `pretest` is because it's easy to forget to rebuild before running `npm test`, and in that case, you may end up chasing down errors that don't exist because you forgot.

xbinaryx: Hmm, should we include both the prepare and pretest scripts in this case?

nzakas: If we include `prepare`, then we should update the [release script](https://github.com/eslint/css/blob/bb90e3a8f08da986817d29ff409bdbf583bfe9df/.github/workflows/release-please.yml#L29) to not run `npm run build` again.

nzakas (approved):
LGTM. Thanks!

PR #192, opened 2025-07-07
fix: enforce strict syntax for `@charset` in no-invalid-at-rules

### LINKED ISSUE #185: Rule Change: Enforce strict syntax for @charset

no-invalid-at-rules

Generate more warnings

A new default behavior

### Example code

```css
@charset 'iso-8859-15'; /* Invalid: single quotes */
@charset  "UTF-8";      /* Invalid: two spaces */
```

Currently, the rule only checks if the prelude is a `<string>`, so both `@charset 'iso-8859-15';` and `@charset  "UTF-8";` are not reported as errors.

After the change, the rule will report an error for any `@charset` that:

- Does not use double quotes for the encoding,
- Does not have exactly one space after `@charset`,
- Is not immediately terminated with a semicolon.

### Participation

### Additional comments

_No response_

thecalamiity:
#### Prerequisites checklist

To enforce strict syntax requirements for the `@charset` rule. This ensures that only valid `@charset` declarations are accepted.

- Updated the no-invalid-at-rules rule to strictly validate `@charset` syntax. Now enforces:
  - The encoding must be enclosed in double quotes.
  - There must be exactly one space after `@charset`.
  - The rule must be immediately terminated with a semicolon.
- Added tests to cover valid and invalid `@charset` usages.
- Updated the documentation to note the strict requirements for `@charset`.

#### Related Issues

Fixes #185

### DIFF
--- a/src/rules/no-invalid-at-rules.js
+++ b/src/rules/no-invalid-at-rules.js
@@ -144,11 +144,21 @@ export default {
 			}
 
 			const nodeText = sourceCode.getText(node);
-			if (!charsetPattern.test(nodeText)) {
-				const preludeText = sourceCode.getText(prelude);
-				const encoding =
-					preludeText.match(charsetEncodingPattern)?.[1] ?? "UTF-8";
+			const preludeText = sourceCode.getText(prelude);
+			const encoding = preludeText
+				.match(charsetEncodingPattern)?.[1]
+				?.trim();
 
+			if (!encoding) {
+				context.report({
+					loc: prelude.loc,
+					messageId: "invalidCharsetSyntax",
+					data: { encoding: "<charset>" },
+				});
+				return;
+			}
+
+			if (!charsetPattern.test(nodeText)) {
 				context.report({
 					loc: prelude.loc,
 					messageId: "invalidCharsetSyntax",
--- a/tests/rules/no-invalid-at-rules.test.js
+++ b/tests/rules/no-invalid-at-rules.test.js
@@ -363,6 +363,32 @@ ruleTester.run("no-invalid-at-rules", rule, {
 			],
 		},
 		{
+			code: '@charset "";',
+			errors: [
+				{
+					messageId: "invalidCharsetSyntax",
+					data: { encoding: "<charset>" },
+					line: 1,
+					column: 10,
+					endLine: 1,
+					endColumn: 12,
+				},
+			],
+		},
+		{
+			code: '@charset "  ";',
+			errors: [
+				{
+					messageId: "invalidCharsetSyntax",
+					data: { encoding: "<charset>" },
+					line: 1,
+					column: 10,
+					endLine: 1,
+					endColumn: 14,
+				},
+			],
+		},
+		{
 			code: "@charset 'UTF-8';",
 			output: '@charset "UTF-8";',
 			errors: [


lumirlumir:
ping @thecalamiity

docs/rules/no-invalid-at-rules.md
```diff
@@ -73,6 +72,29 @@ Examples of **incorrect** code:
 }
 ```
 
+Note on `@charset`: Although it begins with an `@` symbol, it is not an at-rule. It is a specific byte sequence of the following form:
+
+```css
+@charset "<charset>";
+```
+
+where `charset` is a [`<string>`](https://developer.mozilla.org/en-US/docs/Web/CSS/string) denoting the character encoding to be used. It must be the name of a web-safe character encoding defined in the [IANA-registry](https://www.iana.org/assignments/character-sets/character-sets.xhtml), and must be double-quoted, following exactly one space character (U+0020), and immediately terminated with a semicolon.
```
Tanujkanti4441:
```suggestion
where `<charset>` is a [`<string>`](https://developer.mozilla.org/en-US/docs/Web/CSS/string) denoting the character encoding to be used. It must be the name of a web-safe character encoding defined in the [IANA-registry](https://www.iana.org/assignments/character-sets/character-sets.xhtml), and must be double-quoted, following exactly one space character (U+0020) after `@charset`, and immediately terminated with a semicolon.
```

tests/rules/no-invalid-at-rules.test.js
```diff
@@ -293,5 +295,206 @@ ruleTester.run("no-invalid-at-rules", rule, {
...
+						prelude: "'UTF-8'",
+						expected: "<string>",
+					},
+					line: 1,
+					column: 10,
+					endLine: 1,
+					endColumn: 17,
+				},
+			],
+		},
```
Tanujkanti4441: In the case of a syntax error in `@charset`, the error message `invalidPrelude` seems incorrect, such as in this test, the message would be:

```
Invalid prelude ''UTF-8'' found for at-rule '@charset'. Expected '<string>'.
```

Here it says, `<string>` is expected to be used, which is already used here, but with wrong quotes.

And in the case of  `@charset "UTF-8"`, where a semicolon is missing, it has the same error message.

What do you think about adding an error message only to report syntax errors?

thecalamiity: Are you suggesting that instead of reporting `invalidPrelude`, we should report a single, more general error for all these cases? Something like: 
```
invalidCharsetSyntax: "Invalid @charset syntax. Expected: @charset \"<encoding-name>\"; (double quotes, exactly one space after @charset, and a semicolon at the end)."
```

Tanujkanti4441: Yes, but i feel this error message seems a bit descriptive; instead, we can just suggest the correct syntax, like for - 
`@charset "UTF-8"` message can be something like this:

```
Invalid @charset syntax. Expected '@charset "UTF-8";'
```

Please feel free to choose the words that you find more accurate here.

thecalamiity: Just updated the `@charset` error message - PTAL when you can.

src/rules/no-invalid-at-rules.js
```diff
@@ -74,15 +85,91 @@ export default {
...
+		 * @param {AtrulePlain} node The node representing the rule.
+		 */
+		function validateCharsetRule(node) {
+			const { name, prelude, loc } = node;
+
+			const charsetNameLoc = {
+				start: loc.start,
+				end: {
+					line: loc.start.line,
+					column: loc.start.column + name.length + 1,
```
snitin315, line 111: Should be `end`?

```suggestion
				end: {
					line: loc.end.line,
					column: loc.end.column + name.length + 1,
```

src/rules/no-invalid-at-rules.js
```diff
@@ -74,15 +85,91 @@ export default {
...
+					},
+				});
+				return;
+			}
+
+			const nodeText = sourceCode.getText(node);
+			if (!charsetPattern.test(nodeText)) {
+				const preludeText = sourceCode.getText(prelude);
+				const encoding =
+					preludeText.match(charsetEncodingPattern)?.[1] ?? "UTF-8";
```
snitin315: Why do we need `?? "UTF-8"`? In which case will we reach the right operand?

thecalamiity: When the encoding is missing, e.g., `@charset "";`

src/rules/no-invalid-at-rules.js
thecalamiity: Using `start` is correct because we only want to highlight the at-rule name (`@charset`).

snitin315: Shouldn't we report the whole node?

```
@charse "utf-8"
^^^^^^^^^^^^^^^

@charset "";
^^^^^^^^^^^^
```

src/rules/no-invalid-at-rules.js
snitin315: Can we add this test case? empty string.

src/rules/no-invalid-at-rules.js
thecalamiity: We report each issue individually rather than flagging the entire at-rule in one go.
For example, if the name is invalid we report `unknownAtRule`, and if the prelude is missing we report `missingPrelude`, etc.

src/rules/no-invalid-at-rules.js
thecalamiity: I’ve added a test for the empty string case and changed the behavior so we no longer autofix when the encoding is missing or whitespace-only. Previously we defaulted to "UTF-8", but I found this surprising since it guessed a value the user didn’t provide. Now it just reports the issue without autofixing. What do you think of this approach?

snitin315 (approved):
LGTM, thanks! Leaving it open for 2nd review

nzakas (approved):
LGTM. Thanks.

PR #216, opened 2025-07-29
feat: add autofix to no-duplicate-imports rule

### LINKED ISSUE #205: Rule Change: Add autofix capability to the no-duplicate-imports rule

no-duplicate-imports

Implement autofix

Other

### Example code

```css
@import url("reset.css");
@import url("reset.css");
```

Currently, the rule reports an error for the duplicate @import statement, but does not offer an autofix.

After the change, the rule will provide an autofix that removes the duplicate @import rule, resulting in:

```css
@import url("reset.css");
```

### Participation

### Additional comments

_No response_

thecalamiity:
#### Prerequisites checklist

This pull request adds autofix support to the `no-duplicate-imports` rule. The goal is to allow users to automatically remove duplicate `@import` statements in their CSS files.

- Updated the `no-duplicate-imports` rule to provide an autofix that removes duplicate `@import` statements
- Added and updated tests to verify that autofix works correctly

#### Related Issues

Fixes #205

### DIFF
--- a/tests/rules/no-duplicate-imports.test.js
+++ b/tests/rules/no-duplicate-imports.test.js
@@ -129,5 +129,33 @@ ruleTester.run("no-duplicate-imports", rule, {
 				},
 			],
 		},
+		{
+			code: "@import url('a.css');\n@import url('b.css');\n@import url('c.css');\n@import url('a.css');\n@import url('d.css');",
+			output: "@import url('a.css');\n@import url('b.css');\n@import url('c.css');\n@import url('d.css');",
+			errors: [
+				{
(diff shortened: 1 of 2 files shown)


tests/rules/no-duplicate-imports.test.js
nzakas, line 1: Can you add a test where the duplicates are not immediately following each other? Something like this:

```css
@import url(a.css);
@import url(b.css);
@import url(c.css);
@import url(a.css);
@import url(d.css);
```

I think this is a more realistic example of a potential bug.

PR #217, opened 2025-07-29
feat: add suggestion to no-important rule to remove !important flag

### LINKED ISSUE #204: Rule Change: Add autofix capability to the no-important rule

no-important

Implement autofix

Other

### Example code

```css
.button {
  color: red !important;
}
```

Currently, the rule reports an error for the usage of !important, but does not offer an autofix.

After the change, the rule will provide an autofix that removes the !important flag from the declaration, resulting in:

```css
.button {
  color: red;
}
```

### Participation

### Additional comments

_No response_

thecalamiity:
#### Prerequisites checklist

This pull request enhances the `no-important` rule by providing a suggestion to remove the `!important` flag from CSS declarations.

- Added a suggestion to the `no-important` rule that removes the `!important` flag from CSS declarations.
- Updated the test suite to verify that suggestions are provided and that the output matches the expected code with only the `!important` flag removed.

#### Related Issues

Fixes #204

### DIFF
--- a/src/rules/no-important.js
+++ b/src/rules/no-important.js
@@ -16,9 +16,15 @@ import { findOffsets } from "../util.js";
 
 /**
  * @import { CSSRuleDefinition } from "../types.js"
- * @typedef {"unexpectedImportant"} NoImportantMessageIds
+ * @typedef {"unexpectedImportant" | "removeImportant"} NoImportantMessageIds
  * @typedef {CSSRuleDefinition<{ RuleOptions: [], MessageIds: NoImportantMessageIds }>} NoImportantRuleDefinition
  */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const importantPattern = /!(\s|\/\*.*?\*\/)*important/iu;
 
 //-----------------------------------------------------------------------------
 // Rule Definition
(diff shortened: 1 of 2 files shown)


tests/rules/no-important.test.js
```diff
@@ -82,6 +94,12 @@ ruleTester.run("no-important", rule, {
 					column: 34,
 					endLine: 1,
 					endColumn: 45,
+					suggestions: [
+						{
+							messageId: "removeImportant",
+							output: "a { padding: 10px 20px 30px 40px ; }",
```
nzakas: I think when we remove `!important` we should also remove any whitespace before the `!` to avoid these gaps.

tests/rules/no-important.test.js
nzakas, line 1: Can you also add a test case where there is no semicolon after the `!important`? Like this:

```css
a {
    color: red !important
}
```

nzakas (commented):
This looks like a good start. Just a couple of bits to clean up.

src/rules/no-important.js
Tanujkanti4441, line 1: Right now, if somehow the CSS comment before the `!important` flag contains the `!important`, then the rule reports the one inside the comment.

```css
a { color: red /* !important flag 1 */ !important; }


/* after fix the output will be */
a { color: red /*  flag 1 */ !important; }
```

thecalamiity: I opened https://github.com/eslint/css/pull/218 to address this.

PR #218, opened 2025-07-30
fix: prevent false positives for !important inside comments

thecalamiity:
#### Prerequisites checklist

This pull request fixes a bug in the `no-important` rule where `!important` inside CSS comments was incorrectly reported as a violation. The rule should only flag actual `!important` flags in declarations, not those appearing within comments.

- Updated the `no-important` rule to strip comments from the declaration text before searching for `!important`.
- Added tests to verify that `!important` inside comments is ignored, and only real `!important` flags are reported.

#### Related Issues

https://github.com/eslint/css/pull/217#discussion_r2242597429

### DIFF
--- a/tests/rules/no-important.test.js
+++ b/tests/rules/no-important.test.js
@@ -547,6 +547,12 @@ ruleTester.run("no-important", rule, {
 					column: 33,
 					endLine: 1,
 					endColumn: 43,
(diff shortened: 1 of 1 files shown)


nzakas (approved):
LGTM. Thanks!

PR #222, opened 2025-08-01
fix: make relative-font-units unit matching case-insensitive

### LINKED ISSUE #215: Rule Change: Make relative-font-units unit matching case-insensitive

relative-font-units

Generate fewer warnings

A new default behavior

### Example code

```css
/* eslint css/relative-font-units: ["error"] */

.a {
    font-size: 1REM;
}
```

It reports a false positive error, saying that only allowed relative units should be used, even though REM is a valid relative unit (CSS units are case-insensitive).

It will not report an error for 1REM (or any other case variation of allowed units), correctly recognizing that CSS units are case-insensitive and REM is equivalent to rem.

### Participation

### Additional comments

_No response_

xbinaryx:
#### Prerequisites checklist

This PR fixes a false positive error in the `relative-font-units` rule where uppercase CSS units like 1REM, 1EM, etc. were being incorrectly flagged as errors. According to the CSS specification, units are case-insensitive, so REM should be equivalent to rem.

- Modified the `relative-font-units` rule to convert the CSS units to lowercase before comparison
- Added test cases to verify case-insensitive unit matching works correctly

#### Related Issues

Fixes #215

### DIFF
--- a/src/rules/relative-font-units.js
+++ b/src/rules/relative-font-units.js
@@ -103,7 +103,9 @@ export default {
 
 						if (
 							(value.type === "Dimension" &&
-								!allowedFontUnits.includes(value.unit)) ||
+								!allowedFontUnits.includes(
+									value.unit.toLowerCase(),
+								)) ||
 							(value.type === "Identifier" &&
 								disallowedFontSizeKeywords.has(
 									value.name.toLowerCase(),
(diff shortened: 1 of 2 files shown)


nzakas (approved):
LGTM. Thanks.

PR #227, opened 2025-08-04
fix: no-invalid-properties false positives for var() in functions

### LINKED ISSUE #223: Rule Change: no-invalid-properties incorrectly reports for valid CSS with var() in functions

no-invalid-properties

Generate fewer warnings

A new default behavior

### Example code

```css
:root {
  --width: 10px;
}

div {
  width: calc(100% - var(--width))
}
```

The rule incorrectly reports: `Unknown property 'width' found`. This is a regression in the main branch, as the same code works correctly in v0.10.0.

The rule should recognize that `width` is a valid CSS property and that `calc(100% - var(--width))` is a valid value for the width property. It should not report any errors for this code.

### Participation

### Additional comments

_No response_

thecalamiity:
#### Prerequisites checklist

This PR fixes a regression in the `no-invalid-properties` rule where it incorrectly reports for valid CSS properties containing `var()` usage within functions like `calc()`.

- Updated the validation logic to properly handle CSS variables when they appear inside functions.
- Added tests to verify correct behavior for variables in functions

#### Related Issues

Fixes #223


src/rules/no-invalid-properties.js
```diff
@@ -124,6 +124,127 @@ export default {
...
+						!processVarFunction(
+							nestedChild,
+							nestedValueList,
+							valuesWithVarLocs,
+						)
+					) {
+						return null;
+					}
+				} else if (nestedChild.type === "Function") {
+					// Recursively process nested functions
```
nzakas: Here we're doing a traversal on top of the traversal that is already happening in the rule, which is inefficient.

A more efficient approach would be to use the `Function` visitor method to do the traversal automatically. Can you take a look at that?

thecalamiity: I looked into using the Function visitor method here, but I don’t see a straightforward way to apply it. The code at this point isn’t just traversing—it’s reconstructing the property value as a string, including nested function names and arguments. The visitor pattern alone doesn’t provide the context needed for this reconstruction, so a manual traversal seems necessary. If you know of a way to leverage the visitor pattern for this case, I’d appreciate your input.

nzakas: You can start by looking at `Function > *` and `Function > *:exit`. You'd need to use a stack to keep track of the depth you're at, but otherwise, I think that should work.

nzakas:
Can you check the test failures?

tests/rules/no-invalid-properties.test.js
```diff
@@ -55,6 +55,37 @@ ruleTester.run("no-invalid-properties", rule, {
 		":root { --my-color: red; }\na { color: var(--my-color, var(--fallback-color, var(--foo, var(--bar)))) }",
 		":root { --my-color: red; }\na { color: var(--my-color, var(--fallback-color, var(--foo, var(--bar, blue)))) }",
 		":root { --color: red }\na { border-top: 1px var(--style, var(--fallback, solid)) var(--color, blue); }",
+		"a { width: calc(var(--my-width, 100px)) }",
```
snitin315, line 58: Can we add the following test case as well when the variable is also using calc():

```css
/* eslint css/no-invalid-properties: "error" */

:root { --dynamic-width: calc(20px + 10px); }
  
div { width: calc(100% - var(--dynamic-width)); }
```

snitin315 (changes requested):
We have one false positive, can you please check?

```css
/* eslint css/no-invalid-properties: "error" */

:root { 
    --dynamic-width: calc(20px + 10px); 
    --dynamic-width-2: calc(var(--dynamic-width) + 10px); 
}
  
.div { width: calc(100% - var(--dynamic-width)); } // ok
.div2 { width: calc(100% - var(--dynamic-width-2)); } // Error Unknown property 'width' found

```

PR #228, opened 2025-08-04
feat: add allow-list options for at-rules, properties, and selectors

### LINKED ISSUE #210: Rule Change: Add `allowSelectors` option to "use-baseline" rule

use-baseline

Generate fewer warnings

A new option

### Example code

```css
.foo {
  &:hover {
    color: red;
  }
}
```

> Selector 'nesting' is not a widely available baseline feature.

After the change, if a selector is included in `allowSelectors`, it will not report an error even if it's not a baseline feature.

```js
import css from '@eslint/css';

export default {
  files: ['**/*.css'],
  language: 'css/css',
  plugins: { css },
  languageOptions: {
    tolerant: true,
  },
  rules: {
    'css/use-baseline': [
      'error',
      { 
        available: 2021, 
        allowSelectors: ['nesting']  // ✅ New option
      }
    ],
  },
};
```

### Participation

### Additional comments

Our project uses PostCSS to process CSS nesting selector syntax, so the syntax in the Example code is valid for our environment.

We currently cannot set the 'css/use-baseline' rule to 'error' because it would generate numerous false positives in our codebase.

We hope to configure an allowlist of non-baseline selectors in ESLint to prevent errors when using PostCSS features.

I'm uncertain whether similar allow-list options should be supported for other CSS constructs like properties, at-rules, media conditions or types. Currently, I don't know if anyone actually needs these.

Grapedge:
#### Prerequisites checklist

Add new options

- Added `allowAtRules`, `allowProperties`, and `allowSelectors` options
- Added tests and docs

#### Related Issues

fixes #210

### DIFF
--- a/docs/rules/use-baseline.md
+++ b/docs/rules/use-baseline.md
@@ -132,16 +132,27 @@ Examples of **correct** code with `{ allowProperties: ["user-select"] }`:
 
 - `allowSelectors` (default: `[]`) - Specify an array of selectors that are allowed to be used.
 
-Examples of **correct** code with `{ allowSelectors: ["nesting", "has"] }`:
+When you want to allow the [& nesting selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector), you can use `"nesting"`.
+
+Examples of **correct** code with `{ allowSelectors: ["nesting"] }`:
 
 ```css
-/* eslint css/use-baseline: ["error", { allowSelectors: ["nesting", "has"] }] */
+/* eslint css/use-baseline: ["error", { allowSelectors: ["nesting"] }] */
 
-h1 {
-	margin: 0 0 1rem 0;
-	&:has(+ h2) {
-		margin: 0 0 0.25rem 0;
+.parent {
+	&:hover {
+		background-color: blue;
 	}
+}
+```
+
+Examples of **correct** code with `{ allowSelectors: ["has"] }`:
+
+```css
+/* eslint css/use-baseline: ["error", { allowSelectors: ["has"] }] */
+
+h1:has(+ h2) {
+	margin: 0 0 0.25rem 0;
 }
 ```
 
--- a/docs/rules/use-baseline.md
+++ b/docs/rules/use-baseline.md
@@ -99,12 +99,16 @@ Examples of **correct** code:
 
 ### Options
 
-This rule accepts an option object with the following properties:
+This rule accepts an options object with the following properties:
 
 - `available` (default: `"widely"`)
     - change to `"newly"` to allow features that are at the Baseline newly available stage: features that have been supported on all core browsers for less than 30 months
     - set to a numeric baseline year, such as `2023`, to allow features that became Baseline newly available that year, or earlier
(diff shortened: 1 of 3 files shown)


docs/rules/use-baseline.md
```diff
@@ -104,6 +104,46 @@ This rule accepts an option object with the following properties:
...
+```css
+/* eslint css/use-baseline: ["error", { allowSelectors: ["nesting", "has"] }] */
+
+h1 {
+	margin: 0 0 1rem 0;
+	&:has(+ h2) {
+		margin: 0 0 0.25rem 0;
+	}
+}
+```
```
Tanujkanti4441: I think the `nesting` keyword is something we use in the rule, and now it's a part of the option `allowSelector`, so it would be helpful to add a line about what the `nesting` keyword allows?


```css
&:has(+ h2) {
		margin: 0 0 0.25rem 0;
	}
```

Can you also separate the code examples showing the use of `nesting` and `has` in this?

tests/rules/use-baseline.test.js
Tanujkanti4441, line 1: Can you add some `invalid` tests also?

Tanujkanti4441 (commented):
Hi @Grapedge, thanks for the PR. This is looking good, just some suggestions on docs and tests.

docs/rules/use-baseline.md
Grapedge: Thanks for the suggestion.
I've added a brief description of `"nesting"` and I've also split the examples for `nesting` and `has` into two separate cases now.

tests/rules/use-baseline.test.js
Grapedge: Sure, I’ve added it.

docs/rules/use-baseline.md
```diff
@@ -104,6 +104,57 @@ This rule accepts an option object with the following properties:
 - `available` (default: `"widely"`)
     - change to `"newly"` to allow features that are at the Baseline newly available stage: features that have been supported on all core browsers for less than 30 months
     - set to a numeric baseline year, such as `2023`, to allow features that became Baseline newly available that year, or earlier
+- `allowAtRules` (default: `[]`) - Specify an array of at-rules that are allowed to be used.
+
```
snitin315: The general convention is to list all the options initially and then sub headings for each option under . [See this](https://eslint.org/docs/latest/rules/no-underscore-dangle#options) for reference

Grapedge: Thanks for pointing that out! I’ve updated it based on the linked documentation and `use-layers.md`.

tests/rules/use-baseline.test.js
```diff
@@ -364,18 +384,18 @@ ruleTester.run("use-baseline", rule, {
 			],
 		},
 		{
-			code: "@MEDIA (color-gamut: srgb) { a { color: red; } }",
+			code: "@MEDIA (inverted-colors: inverted) { a { color: red; } }",
```
Grapedge: I rebased the main branch and found that I am currently unable to pass the tests. I made modifications by referencing #239.

PR #232, opened 2025-08-08
fix: make no-invalid-properties var() case-insensitive

thecalamiity:
#### Prerequisites checklist

This PR fixes a bug in the `no-invalid-properties` rule where `var()` was matched case-sensitively, causing mixed/uppercase usages like `VAR()` or `Var()` to be missed. CSS function names are ASCII case-insensitive, so these should be treated equivalently.

- Made `var()` detection and fallback parsing case-insensitive
- Added tests for mixed/uppercase `var()` and nested fallbacks

### DIFF
--- a/tests/rules/no-invalid-properties.test.js
+++ b/tests/rules/no-invalid-properties.test.js
@@ -55,6 +55,24 @@ ruleTester.run("no-invalid-properties", rule, {
 		":root { --my-color: red; }\na { color: var(--my-color, var(--fallback-color, var(--foo, var(--bar)))) }",
 		":root { --my-color: red; }\na { color: var(--my-color, var(--fallback-color, var(--foo, var(--bar, blue)))) }",
 		":root { --color: red }\na { border-top: 1px var(--style, var(--fallback, solid)) var(--color, blue); }",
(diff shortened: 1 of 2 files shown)


tests/rules/no-invalid-properties.test.js
```diff
@@ -592,5 +618,134 @@ ruleTester.run("no-invalid-properties", rule, {
 				},
 			],
 		},
+		{
```
snitin315, line 671: CSS variables `--my-color` is not case sensitive, can we please add a few cases to verify the rule is working as expected for them?

thecalamiity: They are case-sensitive. From [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties):

> Note: Custom property names are case sensitive — `--my-color` will be treated as a separate custom property to `--My-color`.

snitin315: Ah sorry! I meant case sensitive only, just add a case where it is working as expected

snitin315 (approved):
LGTM, thanks!

PR #233, opened 2025-08-09
fix: treat at-rule names as case-insensitive across rules

thecalamiity:
#### Prerequisites checklist

Ensure at-rule names are treated case-insensitively across rules, aligning with CSS syntax and preventing missed detections when authors use mixed-case at-rules.

- Updated rule listeners to use case-insensitive matching
- Added mixed-case variants to tests to verify case-insensitive behavior

### DIFF
--- a/src/rules/no-duplicate-imports.js
+++ b/src/rules/no-duplicate-imports.js
@@ -42,7 +42,7 @@ export default {
 		const imports = new Set();
 
 		return {
-			"Atrule[name=import]"(node) {
+			"Atrule[name=/^import$/i]"(node) {
 				const url = node.prelude.children[0].value;
 
 				if (imports.has(url)) {
--- a/src/rules/no-duplicate-keyframe-selectors.js
+++ b/src/rules/no-duplicate-keyframe-selectors.js
@@ -39,12 +39,12 @@ export default {
 		const seen = new Map();
 
 		return {
-			"Atrule[name=keyframes]"() {
+			"Atrule[name=/^keyframes$/i]"() {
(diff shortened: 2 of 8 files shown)


src/rules/no-duplicate-imports.js
```diff
@@ -42,7 +42,7 @@ export default {
 		const imports = new Set();
 
 		return {
-			"Atrule[name=import]"(node) {
+			"Atrule[name=/^import$/i]"(node) {
```
snitin315, line 45: Can you also share the official reference where it is mentioned these are case insensitive?

thecalamiity: The at-rules affected in this PR are defined across different specs, so I’ll need to track down the relevant references. In the meantime, here’s a [StackBlitz](https://stackblitz.com/edit/web-platform-7txzkjm7?file=a.css) reproduction for `@import`.

nzakas:
@thecalamiity it appears you are correct. I had to dig deep into the [spec](https://www.w3.org/TR/css-syntax-3/#rule-defs) to find this:

> Tokens match if their value is a match for the value defined in the grammar. Unless otherwise specified, all matches are [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive).

In the future, keep in mind that simple repros in a single browser don't mean much -- browsers can implement things different ways. What is more useful is finding the text in the specification that backs up your claim.

nzakas (approved):
LGTM. Would like @snitin315 to review before merging.

PR #252, opened 2025-09-03
feat: add `selector-complexity` rule

### LINKED ISSUE #19: New Rule: selector-complexity

Tanujkanti4441:
#### Prerequisites checklist

To add a new rule to limit CSS selectors.

Added a new rule called `selector-complexity` with following options:

`maxIds`
`maxClasses`
`maxTypes`
`maxAttributes`
`maxPseudoClasses`
`maxUniversals`
`maxCompounds`
`maxCombinators`
`disallowCombinators`
`disallowPseudoClasses`
`disallowPseudoElements`
`disallowAttributes`
`disallowAttributeMatchers`

#### Related Issues
Fixes #19

### DIFF
--- a/src/rules/selector-complexity.js
+++ b/src/rules/selector-complexity.js
@@ -33,37 +33,10 @@
 //-----------------------------------------------------------------------------
 
 /**
- * Extract compounds from a CSS selector.
- * @param {Array<Object>} selectors All CSS selector nodes.
- * @returns {Array<Array<Object>>} Array of compounds (each compound is an array of selector nodes).
- */
-function getCompounds(selectors) {
-	const compounds = [];
-	let compound = [];
-
-	selectors.forEach(selector => {
-		if (selector.type === "Combinator") {
-			if (compound.length > 0) {
-				compounds.push(compound);
-				compound = [];
-			}
-		} else {
-			compound.push(selector);
-		}
-	});
-
-	if (compound.length > 0) {
-		compounds.push(compound);
-	}
-
-	return compounds;
-}
-
-/**
  * Get the location of a selector of a given name.
  * @param {Array<Object>} allSelector All CSS selector nodes.
  * @param {string} disallowedSelector The name of the disallowed selector.
- * @returns {Object | undefined} The location of the disallowed selector, or undefined if not found.
+ * @returns {Object} The location of the disallowed selector.
  */
 function getDisallowedSelectorsLocation(allSelector, disallowedSelector) {
 	return allSelector.find(selector => selector.name === disallowedSelector)
@@ -75,6 +48,7 @@ function getDisallowedSelectorsLocation(allSelector, disallowedSelector) {
  * @param {Object} context The ESLint rule context object.
  * @param {Object} selectorLoc The location of the selector.
  * @param {number} maxValue The max number of selectors that are allowed.
+ * @param {string} selectorType The type of CSS selector.
  * @returns {void}
  */
 function exceedLimitError(context, selectorLoc, maxValue, selectorType) {
@@ -111,7 +85,7 @@ function getSelectorNames(selectors) {
  * Get the location of the attribute matcher or operator in a given attribute selector.
  * @param {Array<Object>} selectors All CSS selector nodes.
  * @param {number} index The index of the attribute selector in the selectors array.
- * @returns {{ startLoc: Object | undefined, endLoc: Object | undefined }} The start and end locations of the operator.
+ * @returns {{ startLoc: Object, endLoc: Object }} The start and end locations of the operator.
  */
 function getOperatorLocation(selectors, index) {
 	const selector = selectors[index];
@@ -135,7 +109,7 @@ function getOperatorLocation(selectors, index) {
  * @param {Array<Object>} combinatorNodes All combinator nodes.
  * @param {string} combinator Name of combinator.
  * @param {number} index The index of the given combinator.
- * @returns {Object | undefined} The location of the disallowed combinator, or undefined if not found.
+ * @returns {Object} The location of the disallowed combinator.
  */
 function getDisallowedCombinatorsLocation(
 	selectors,
@@ -333,7 +307,6 @@ export default {
 						child.type === "TypeSelector" && child.name === "*",
 				);
 				const combinatorNodes = getSelectors(selectors, "Combinator");
-				const compounds = getCompounds(selectors);
 
 				const combinators = getSelectorNames(combinatorNodes);
 				const pseudoClassSelectorsNames =
@@ -398,7 +371,7 @@ export default {
 					);
 				}
 
-				if (compounds.length > maxCompounds) {
+				if (combinatorNodes.length + 1 > maxCompounds) {
 					exceedLimitError(
 						context,
 						selectorLoc,


src/rules/selector-complexity.js
```diff
@@ -0,0 +1,488 @@
...
+/**
+ * An error for exceeding the maximum allowed selectors of a specific type.
+ * @param {Object} context The ESLint rule context object.
+ * @param {Array<Object>} selectorsList List of CSS selectors.
+ * @param {number} maxValue The max number of selectors that are allowed.
+ * @returns {void}
+ */
+function exceedLimitError(context, selectorsList, maxValue, selectorType) {
+	context.report({
+		loc: selectorsList[maxValue].loc,
```
fasttime: For a descendant combinator the `loc` property is `null`, which is not allowed by `context.report`.

For example, this test case is failing with an error:

```js
{
	code: ".foo .bar {}",
	options: [{ maxCombinators: 0 }],
	errors: 1,
}
```

```text
  1) selector-complexity
       invalid
         .foo .bar {}:
     Error: Node must be provided when reporting error if location is not provided
Occurred while linting <input>:1
Rule: "rule-to-test/selector-complexity"
      at ok (node_modules/eslint/lib/shared/assert.js:17:9)
      at assertValidNodeInfo (node_modules/eslint/lib/linter/file-report.js:173:3)
      at FileReport.addRuleMessage (node_modules/eslint/lib/linter/file-report.js:538:3)
      at FileContext.report (node_modules/eslint/lib/linter/linter.js:1111:28)
      at exceedLimitError (file:///.../css/src/rules/selector-complexity.js:81:10)
      at Selector (file:///.../css/src/rules/selector-complexity.js:352:6)
      at ruleErrorHandler (node_modules/eslint/lib/linter/linter.js:1173:33)
      at /.../css/node_modules/eslint/lib/linter/source-code-visitor.js:76:46
      at Array.forEach (<anonymous>)
      at SourceCodeVisitor.callSync (node_modules/eslint/lib/linter/source-code-visitor.js:76:30)
      at /.../css/node_modules/eslint/lib/linter/source-code-traverser.js:291:18
      at Array.forEach (<anonymous>)
      at SourceCodeTraverser.traverseSync (node_modules/eslint/lib/linter/source-code-traverser.js:290:10)
      at runRules (node_modules/eslint/lib/linter/linter.js:1214:12)
      at #flatVerifyWithoutProcessors (node_modules/eslint/lib/linter/linter.js:2101:4)
      at Linter._verifyWithFlatConfigArrayAndWithoutProcessors (node_modules/eslint/lib/linter/linter.js:2189:43)
      at Linter._verifyWithFlatConfigArray (node_modules/eslint/lib/linter/linter.js:2292:15)
      at Linter.verify (node_modules/eslint/lib/linter/linter.js:1677:10)
      at runRuleForItem (node_modules/eslint/lib/rule-tester/rule-tester.js:884:23)
      at testInvalidTemplate (node_modules/eslint/lib/rule-tester/rule-tester.js:1075:19)
      at Context.<anonymous> (node_modules/eslint/lib/rule-tester/rule-tester.js:1564:10)
      at process.processImmediate (node:internal/timers:485:21)
```

I think in this and other cases it would be better to report on the whole selector (`.foo .bar`).

fasttime (commented):
Excited to see this rule taking shape. I noticed that currently the rule only reports on a single node in the selector:

Image

This is different from Stylelint, which reports on the whole selector (`#foo #bar #baz`) when a it finds a problem:

[**Stylelint Demo**](https://stylelint.io/demo/#N4Igxg9gJgpiBcIDEAzCEAESBGBDATlngF4bAA6AduQC40wAeNAtLgDYCWA5pfBmDEr18AbioBfKiAA0IFBzYwAcrgC2cRIzUAHRQDowAZ0MzwESvK4IQFShgzkQ+AK6LDjvrfv3HhmIrAaCHxmVVwGZg4oDwwARip7SUpxU0gLbgAxYLCaawArQ3NTWG0TRC8HEEMaAE9FTiEYxzZcemrHCRBxIA)

I believe that reporting on the whole selector would be less confusing because it's not obvious which part of a selector should be changed to comply with the rule. Thoughts?

src/rules/selector-complexity.js
Tanujkanti4441: Done! The rule will now report the entire selector when the selector limit is exceeded.

src/rules/selector-complexity.js
```diff
@@ -0,0 +1,479 @@
...
+
+				if (disallowCombinators.length > 0) {
+					let disallowedCombinatorLocation;
+					for (const combinator of combinators) {
+						if (disallowCombinators.includes(combinator)) {
+							disallowedCombinatorLocation =
+								getDisallowedSelectorsLocation(
+									combinatorNodes,
+									combinator,
+								);
```
fasttime, line 414: When `combinator` is `" "`, `disallowedCombinatorLocation` will be `null`. This is because of a restriction of `css-tree`, since the combinator has no tokens. As a result, `RuleTester` will throw an error. For example, this test case:

```js
{
    only: true,
    code: "foo bar {}",
    options: [{ disallowCombinators: [" "] }],
    errors: [
        {
            messageId: "disallowedSelectors",
            data: {
                selectorName: " ",
                selector: "combinator",
            }
        },
    ],
}
```

Is currently failing with:

```text
     Error: Node must be provided when reporting error if location is not provided
Occurred while linting <input>:1
Rule: "rule-to-test/selector-complexity"
```

There are different possible ways to address this:

1. Report the location of the whole selector string.
2. Report the start of the selector string as a location.
3. Filter out the `" "` combinator in the `disallowCombinators` array to avoid reporting it. We could say that the descendant combinator cannot be disallowed.

I think reporting any location would be still more useful than ignoring the combinator altogether. Thoughts?

src/rules/selector-complexity.js
```diff
@@ -0,0 +1,479 @@
...
+			Selector(node) {
+				const selectors = node.children;
+				const selectorLoc = node.loc;
+
+				const idSelectors = getSelectors(selectors, "IdSelector");
+				const classSelectors = getSelectors(selectors, "ClassSelector");
+				const typeSelectors = selectors.filter(
+					child =>
+						child.type === "TypeSelector" && child.name !== "*",
+				);
```
fasttime, line 296: Can we add a test to verify that `*` nodes are not counted as type selectors?

src/rules/selector-complexity.js
```diff
@@ -0,0 +1,479 @@
...
+					"AttributeSelector",
+				);
+				const pseudoClassSelectors = getSelectors(
+					selectors,
+					"PseudoClassSelector",
+				);
+				const universalSelectors = selectors.filter(
+					child =>
+						child.type === "TypeSelector" && child.name === "*",
+				);
```
fasttime, line 308: Conversely, we could add a test to ensure that regular `TypeSelector`s aren't counted as universals.

src/rules/selector-complexity.js
Tanujkanti4441: How about if we use the end and start location of the previous and next node of the `" "` combinator respectively, as we are doing with attribute operators here?

fasttime: If there is a good way to locate the previous and next node of the combinator, that would be perfect 👍 From what I can see, the combinator nodes are deeply equal, so I guess they can be only identified by object reference.

src/rules/selector-complexity.js
Tanujkanti4441: Done!

src/rules/selector-complexity.js
fasttime: Nice!

src/rules/selector-complexity.js
```diff
@@ -0,0 +1,526 @@
...
+						index,
+						attributeMatcher,
+					] of attributeMatchers.entries()) {
+						if (
+							disallowAttributeMatchers.includes(attributeMatcher)
+						) {
+							const { startLoc, endLoc } = getOperatorLocation(
+								attributeSelectors,
+								index,
+							);
```
fasttime, line 480: The location for attribute matchers is incorrect sometimes:

```js
/* eslint css/selector-complexity: ["error", { disallowAttributeMatchers: ["*="] }] */

[foo] [bar*="baz"] {
  color: black;
}
```

image

Tanujkanti4441: Done!

src/rules/selector-complexity.js
```diff
@@ -0,0 +1,526 @@
...
+		.loc;
+}
+
+/**
+ * An error for exceeding the maximum allowed selectors of a specific type.
+ * @param {Object} context The ESLint rule context object.
+ * @param {Object} selectorLoc The location of the selector.
+ * @param {number} maxValue The max number of selectors that are allowed.
+ * @returns {void}
+ */
```
fasttime, line 53: The JSDoc `@param` entry for `selectorType` is missing.

src/rules/selector-complexity.js
```diff
@@ -0,0 +1,526 @@
...
+	return { startLoc, endLoc };
+}
+
+/**
+ * Get the location of a given disallowed combinator.
+ * @param {Array<Object>} selectors All CSS selector nodes.
+ * @param {Array<Object>} combinatorNodes All combinator nodes.
+ * @param {string} combinator Name of combinator.
+ * @param {number} index The index of the given combinator.
+ * @returns {Object | undefined} The location of the disallowed combinator, or undefined if not found.
```
fasttime: The return value should not be `undefined` I think, otherwise an exception will be thrown.

src/rules/selector-complexity.js
```diff
@@ -0,0 +1,526 @@
...
+				if (combinatorNodes.length > maxCombinators) {
+					exceedLimitError(
+						context,
+						selectorLoc,
+						maxCombinators,
+						"combinator",
+					);
+				}
+
+				if (compounds.length > maxCompounds) {
```
fasttime: I think by the current logic the number of compounds can be calculated by adding 1 to the number of combinators, making the `getCompounds` function unnecessary.

```suggestion
				if (combinatorNodes.length + 1 > maxCompounds) {
```

fasttime (approved):
LGTM, thanks! Leaving open for a second review before merging.

PR #297, opened 2025-10-13
fix: correct message for functions in `use-baseline` rule

### LINKED ISSUE #293: Rule Change: Incorrect error message for CSS functions in `use-baseline` rule

use-baseline

Other

Other

### Example code

```css
.a {
  color: light-dark(#333b3c, #efefec);
}
```

The rule reports:

```
Type 'light-dark' is not a widely available baseline feature.
```

This is confusing because `light-dark` is clearly a CSS function, not a "type". The term "type" is misleading and doesn't accurately describe what's being checked.

After the change, the rule will report:

```
Function 'light-dark' is not a widely available baseline feature.
```

### Participation

### Additional comments

The data generation script already correctly filters to only include CSS functions (by checking `mdnData.css.functions`), but currently exports them under the misleading name `types`. We should update the data export name to `functions` for consistency.

thecalamiity:
#### Prerequisites checklist

This pull request fixes an inaccurate error message in the `use-baseline` rule. Previously, CSS functions (such as `light-dark()` and `color-mix()`) were reported as “types,” which was misleading. The rule will now correctly refer to them as “functions.”

- Renamed the exported baseline data key from `types` → `functions`.  
- Updated the `use-baseline` rule to:
  - Use `functions` instead of `types`.  
  - Replace the message ID `notBaselineType` with `notBaselineFunction`.  
  - Update the corresponding error message to use “Function” instead of “Type.”  
- Updated the baseline data generation script to export `functions` consistently.  
- Adjusted all affected test cases to match the new message ID and terminology.

#### Related Issues

Fixes #293

### DIFF
--- a/tools/generate-baseline.js
+++ b/tools/generate-baseline.js
@@ -109,7 +109,7 @@ function extractCSSFeatures(features) {
 	const propertyValues = {};
 	const atRules = {};
 	const mediaConditions = {};
-	const types = {};
+	const functions = {};
 	const selectors = {};
 
 	for (const [key, featureId] of Object.entries(features)) {
@@ -152,13 +152,13 @@ function extractCSSFeatures(features) {
 			continue;
 		}
 
-		// types
+		// functions
 		if ((match = cssTypePattern.exec(key)) !== null) {
 			const type = match.groups.type;
 			if (!(`${type}()` in mdnData.css.functions)) {
 				continue;
 			}
-			types[type] = mapFeatureStatus(status);
+			functions[type] = mapFeatureStatus(status);
 			continue;
 		}
 
@@ -174,7 +174,7 @@ function extractCSSFeatures(features) {
 		propertyValues,
 		atRules,
 		mediaConditions,
-		types,
+		functions,
 		selectors,
 	};
 }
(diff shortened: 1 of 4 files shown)


tools/generate-baseline.js
```diff
@@ -210,7 +210,7 @@ export const BASELINE_FALSE = ${BASELINE_FALSE};
 export const properties = new Map(${JSON.stringify(Object.entries(cssFeatures.properties), null, "\t")});
 export const atRules = new Map(${JSON.stringify(Object.entries(cssFeatures.atRules), null, "\t")});
 export const mediaConditions = new Map(${JSON.stringify(Object.entries(cssFeatures.mediaConditions), null, "\t")});
-export const types = new Map(${JSON.stringify(Object.entries(cssFeatures.types), null, "\t")});
+export const functions = new Map(${JSON.stringify(Object.entries(cssFeatures.functions), null, "\t")});
```
nzakas, line 213: I don't think this is the correct approach here. Let's keep `types` as-is and just add `functions`.

thecalamiity: The "types" here are actually just functions. We're filtering the `css.types.*` entries through `mdnData.css.functions` to extract only the ones that represent actual CSS functions.

https://github.com/eslint/css/blob/c6c5b9ddbcde22b4c65381dce82c770de44e8b1e/tools/generate-baseline.js#L158-L160

nzakas: Ah gotcha, thanks for explaining. :+1:

Issue #186, opened 2025-07-02
New Rule: no-unmatchable-anb-selector


TKDev7:
### Rule details

Disallow unmatchable An+B selectors.

### What type of rule is this?

Warns about a potential problem

### Example code

```css
a:nth-child(0n+0) {}
/*↑             ↑
 * This unmatchable An+B selector */
```

### Prior Art

```markdown
selector-anb-no-unmatchable: https://stylelint.io/user-guide/rules/selector-anb-no-unmatchable/
```

### Participation

- [x] I am willing to submit a pull request to implement this rule.

### Additional comments

[An+B selectors](https://www.w3.org/TR/css-syntax-3/#anb-microsyntax) are one-indexed. Selectors that always evaluate to 0 will not match any elements.

nzakas:
Sounds good to me. Marking as accepted. :+1:

nzakas:
> Is it worth expanding the scope of this rule to detect more unmatchable selectors and naming it `no-unmatchable-selectors`?

I like that idea. It also gives us flexibility to start with a small group that we're sure of and add more over time.

@thecalamiity what do you think?

Mouvedia:
> …but support for detecting them has been requested in Stylelint…

stylelint/stylelint#8566
stylelint/stylelint#5575

snitin315:
@TKDev7 Are you still working on this?

Tanujkanti4441:
Hey @TKDev7, are you still working on this, or would you like me to take it up?

PR #301, opened 2025-10-14
feat: add `no-unmatchable-selectors` rule

TKDev7:
#### Prerequisites checklist

This PR introduces a new rule `no-unmatchable-selectors` to catch CSS selectors that can never match any elements, helping developers identify dead code and potential mistakes in their stylesheets.

- Implemented `no-unmatchable-selectors` rule that detects and reports selectors with impossible `:nth-*()` pseudo-class formulas
- Added rule documentation
-  Included test cases covering both invalid patterns and valid patterns

#### Related Issues

#186

This initial implementation focuses on `:nth-*()` pseudo-classes, but I plan to expand the rule in follow-up PRs to cover additional scenarios.

### DIFF
--- a/docs/rules/no-unmatchable-selectors.md
+++ b/docs/rules/no-unmatchable-selectors.md
@@ -6,6 +6,11 @@ Disallow unmatchable selectors.
 
 An unmatchable selector is one that can never match any element in any document. These are effectively dead code and usually indicate mistakes.
 
+For example:
+
+- `a:nth-child(0)` — the `An+B` formula never produces a positive position (≥ 1).
+- `a:nth-child(-n)` — a negative step with no offset never yields a positive position.
+
 ## Rule Details
 
 This rule reports selectors that can never match any element.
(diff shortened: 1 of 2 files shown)


docs/rules/no-unmatchable-selectors.md
```diff
@@ -0,0 +1,50 @@
+# no-unmatchable-selectors
+
+Disallow unmatchable selectors.
+
+## Background
+
+An unmatchable selector is one that can never match any element in any document. These are effectively dead code and usually indicate mistakes.
```
nzakas, line 7: Let's give a couple of examples here.

src/rules/no-unmatchable-selectors.js
```diff
@@ -0,0 +1,59 @@
...
+		},
+	},
+
+	create(context) {
+		const { sourceCode } = context;
+
+		return {
+			AnPlusB(node) {
+				const a = Number(node.a);
+				const b = Number(node.b);
```
nzakas, line 45: Can you add a comment here describing what's happening? Either `node.a` or `node.b` can be `null`, in which case `Number()` converts it into 0. I don't think it's obvious from the code that this is an intentional behavior and not an oversight.

PR #310, opened 2025-10-27
feat: add `allow*` options to `use-baseline` rule

### LINKED ISSUE #298: Rule Change: Add more `allow*` options to `use-baseline` rule

use-baseline

Generate fewer warnings

Add options to allow specific CSS functions, media conditions, and property values to be exempt from baseline enforcement.

### Example code

```css
.selector {
  color: color-mix(in srgb, plum, #123456);
  clip-path: fill-box;
}

@media (scripting: none) {}
```

It reports that:
- The `color-mix()` function is not a widely available baseline feature.
- The value `fill-box` for `clip-path` is not a widely available baseline feature.
- The media condition (`scripting: none`) is not a widely available baseline feature.

Currently, these cannot be selectively allowed.

If configured like this:

```js
"css/use-baseline": [
  "error",
  {
    "available": "widely",
    "allowFunctions": ["color-mix"],
    "allowMediaConditions": ["scripting"],
    "allowPropertyValues": {
      "clip-path": ["fill-box"]
    }
  }
]
```

then the rule would not report errors for the above code.

### Participation

### Additional comments

_No response_

thecalamiity:
#### Prerequisites checklist

This PR adds more flexible configuration to the `use-baseline` rule by introducing the `allowFunctions`, `allowMediaConditions`, and `allowPropertyValues` options. These options let users selectively permit specific functions, media conditions, and property values that are otherwise reported as non-baseline features.

- Added `allowFunctions`, `allowMediaConditions`, and `allowPropertyValues` options to the rule schema and implementation.
- Updated documentation with examples.
- Added corresponding tests.

#### Related Issues

Fixes #298

### DIFF
--- a/src/rules/use-baseline.js
+++ b/src/rules/use-baseline.js
@@ -439,45 +439,53 @@ export default {
 					allowAtRules: {
 						type: "array",
 						items: {
-							type: "string",
+							enum: Array.from(atRules.keys()),
 						},
 						uniqueItems: true,
 					},
 					allowFunctions: {
 						type: "array",
 						items: {
-							type: "string",
+							enum: Array.from(functions.keys()),
 						},
 						uniqueItems: true,
 					},
 					allowMediaConditions: {
 						type: "array",
 						items: {
-							type: "string",
+							enum: Array.from(mediaConditions.keys()),
 						},
 						uniqueItems: true,
 					},
 					allowProperties: {
 						type: "array",
 						items: {
-							type: "string",
+							enum: Array.from(properties.keys()),
 						},
 						uniqueItems: true,
 					},
 					allowPropertyValues: {
 						type: "object",
-						additionalProperties: {
-							type: "array",
-							items: {
-								type: "string",
-							},
-							uniqueItems: true,
-						},
+						properties: Object.fromEntries(
+							Array.from(propertyValues.entries()).map(
+								([prop, valuesMap]) => [
+									prop,
+									{
+										type: "array",
+										items: {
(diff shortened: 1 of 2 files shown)


src/rules/use-baseline.js
```diff
@@ -440,13 +443,37 @@ export default {
...
+						},
+						uniqueItems: true,
+					},
+					allowMediaConditions: {
+						type: "array",
+						items: {
+							type: "string",
+						},
+						uniqueItems: true,
+					},
```
Tanujkanti4441, line 459: I think we can also validate these new and existing options with baseline data?

tests/rules/use-baseline.test.js
```diff
@@ -604,5 +627,87 @@ ruleTester.run("use-baseline", rule, {
...
+					},
+					line: 1,
+					column: 9,
+					endLine: 1,
+					endColumn: 24,
+				},
+			],
+		},
+		{
+			code: "@supports (clip-path: fill-box) { a { clip-path: stroke-box; } }",
```
Tanujkanti4441, line 669: Can you also add such example in `valid` where `clip-path: fill-box` is allowed by `@supports` and `clip-path: stroke-box` is allowed by allowPropertyValues option?

src/rules/use-baseline.js
thecalamiity: Yes 👍🏻

Tanujkanti4441 (approved):
LGTM, thanks!

Would like another review before merging.

PR #313, opened 2025-10-29
fix: correct parent typing for rule visitors

crimsonjay0:
#### Prerequisites checklist

This PR updates `CSSNodeVisitor` so non-root nodes include a `parent` parameter; for root, only the `node` is passed.

Updated `CSSNodeVisitor` so non-root nodes include a `parent` parameter; for root, only the node is passed.

### DIFF
--- a/tests/types/types.test.ts
+++ b/tests/types/types.test.ts
@@ -167,8 +167,10 @@ css.configs.recommended.plugins satisfies object;
 			Function: (...args) => testVisitor<FunctionNodePlain>(...args),
 			"Function:exit": (...args) =>
 				testVisitor<FunctionNodePlain>(...args),
-			GeneralEnclosed: (...args) => testVisitor<any>(...args),
-			"GeneralEnclosed:exit": (...args) => testVisitor<any>(...args),
+			GeneralEnclosed: (node: any, parent: CssNodePlain) =>
+				testVisitor<any>(node, parent),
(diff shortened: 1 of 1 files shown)


tests/types/types.test.ts
```diff
@@ -114,115 +117,137 @@ css.configs.recommended.plugins satisfies object;
...
+				testVisitor<UnicodeRange>(...args),
+			Url: (...args) => testVisitor<Url>(...args),
+			"Url:exit": (...args) => testVisitor<Url>(...args),
+			Value: (...args) => testVisitor<ValuePlain>(...args),
+			"Value:exit": (...args) => testVisitor<ValuePlain>(...args),
+			WhiteSpace: (...args) => testVisitor<WhiteSpace>(...args),
+			"WhiteSpace:exit": (...args) => testVisitor<WhiteSpace>(...args),
+
+			// Unknown selectors allowed
+			"Atrule[name=import]"(node: AtrulePlain, parent: CssNodePlain) {},
```
DMartens, line 254: The other changes LGTM, thanks.
This is not an "unknown" selector but rather a combined selector.
Can you please update the comment and also add test cases for unknown selectors (e.g. ForStatement, Unknown, ValueNode) and for additional combined selectors (e.g. "Value, Url)", "*", "Block:first-child")

DMartens (approved):
Thank for adding the test cases, changes LGTM.

PR #330, opened 2025-11-18
build: Ensure auto-created PRs run CI

nzakas:
#### Prerequisites checklist

To make the update-baseline workflow run CI when opened.

I switched the workflow to use our GitHub Bot app to create the PR instead of the default workflow user. This should trigger the other workflows to run correctly.

**Note:** `BOT_APP_PRIVATE_KEY` is only enabled for this repo. To do the same thing in other repos, we'd need to add access for those repos.

Reference:
https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

### DIFF
--- a/.github/workflows/update-baseline.yml
+++ b/.github/workflows/update-baseline.yml
@@ -6,21 +6,8 @@ on:
 
 jobs:
     update-baseline:
-        env:
-            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
         runs-on: ubuntu-latest
         steps:
-            # We are using our GitHub Bot App to create the PR.
-            # This ensures that other workflows triggered by
-            # PR creation will run as expected.
-            # The GitHub Bot App is configured in the organization settings.
-            - uses: actions/create-github-app-token@v2
-              id: generate-token
-              with:
-                  app-id: ${{ vars.BOT_APP_ID }}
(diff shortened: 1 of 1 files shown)


lumirlumir (changes requested):
Thanks for taking a look at this issue!

I investigated and manually triggered the workflow on the `auto-pr-fix` branch, but it encountered some errors:

https://github.com/eslint/css/actions/runs/19489195700/job/55777962002

Image

It seems there are errors like the following:

- `Inputs 'owner' and 'repositories' are not set.`
- `Failed to create token for "css" (attempt 1): [universal-github-app-jwt] Private Key is in OpenSSH format, but only PKCS#8 is supported.`

---

For other reviewers' convenience, here are some additional references:

- Authenticating with GitHub App generated tokens: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens
- Creating pull requests in a remote repository using GitHub App generated tokens: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#creating-pull-requests-in-a-remote-repository-using-github-app-generated-tokens
- `create-github-app-token` Action: https://github.com/actions/create-github-app-token?tab=readme-ov-file#create-github-app-token

PR #337, opened 2025-12-02
fix: update `@eslint/core` to `v1.0.0` and adjust tests

lumirlumir:
## Prerequisites checklist

## What is the purpose of this pull request?

This PR was motivated by https://github.com/eslint/json/pull/187#issuecomment-3602811830.

Currently, when updating `@eslint/core` to v1.0.0, CI fails because `@eslint/core@1.0.0` does not match ESLint v9

`@eslint/core@1.0.0` includes type information for ESLint v10, which causes the type mismatch.

So, I've updated ESLint to `>=10.0.0-alpha.0 <10.0.0 || ^10.0.1`. (Ref: https://github.com/eslint/eslint/blob/main/packages/eslint-config-eslint/package.json#L66)

---

When I update ESLint to `>=10.0.0-alpha.0 <10.0.0 || ^10.0.1`, the following error occurs, so I added the `legacy-peer-deps` option to `.npmrc`.

This change follows the approach in https://github.com/eslint/eslint/pull/20281

image

## What changes did you make? (Give an overview)

In this PR, I've updated `@eslint/core`, add `legacy-peer-deps`, and use ESLint v10 pre-release.

## Related Issues

Ref: #327, https://github.com/eslint/json/pull/187#issuecomment-3602811830, https://github.com/eslint/eslint/pull/20281

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
   ],
   "license": "Apache-2.0",
   "dependencies": {
-    "@eslint/core": "^1.0.0",
+    "@eslint/core": "^1.0.1",
     "@eslint/css-tree": "^3.6.8",
     "@eslint/plugin-kit": "^0.5.1"
   },


package.json
```diff
@@ -84,15 +84,15 @@
   ],
   "license": "Apache-2.0",
   "dependencies": {
-    "@eslint/core": "^0.17.0",
+    "@eslint/core": "^1.0.0",
```
mdjermanovic: Same question as https://github.com/eslint/json/pull/190#discussion_r2584390444. We can clarify there, just wanted to leave a note here as well so that we don't merge until this is clarified.

nzakas (approved):
LGTM. Thanks.

PR #371, opened 2026-01-29
chore: add lint:unused script to run Knip

Pixel998:
#### Prerequisites checklist

Adds [Knip](https://knip.dev) to check for unused exports, unused dependencies, and missing dependencies

- Added `knip.json` config and a root `lint:unused` script.
- Added a "Lint Files, Dependencies, & Exports" step to our CI.
- Removed unused `BASELINE_FALSE` export

#### Related Issues

eslint/js#695

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "eslint-config-eslint": "^14.0.0",
     "eslint-plugin-eslint-plugin": "^7.3.2",
     "globals": "^17.0.0",
-    "knip": "^5.82.1",
+    "knip": "^5.87.0",
     "lint-staged": "^16.0.0",
     "mdast-util-from-markdown": "^2.0.2",
     "mdn-data": "^2.27.1",


knip.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "entry": ["tests/**/*.{cts,js,ts}"]
```
DMartens: Knip does not seem to recognize the glob pattern in the package.json test script.
The [knip documentation](https://knip.dev/guides/configuring-project-files#production-mode) advices to use `--production` instead of ignoring the test files.

Pixel998: Sorry, are you suggesting we change something here?

DMartens: I did not mean "ignoring" but removing the "tests/**/*.{cts,js,ts}" pattern from entry (as mentioned in the linked documentation in the second example).

nzakas (approved):
LGTM. Would like @DMartens to verify his feedback has been addressed before merging.

knip.json
Pixel998: The [Mocha](https://knip.dev/reference/plugins/mocha) plugin defaults to `**/test/*.{js,cjs,mjs}`, so it won’t discover our `tests/**/*.test.js` layout by default. Without an explicit entry, the entire `tests/` tree is reported as unused and test-only dev deps (e.g. `dedent`, `tailwind-csstree`) get flagged.

On production mode: Knip’s docs are clear that `--production` doesn’t replace the default run; it’s an additional, shipped-code-only view. I think it can be useful to run both. That said, we don’t currently run the production pass in our other repos, so it’s probably better to add `knip --production` as a follow-up PR once we’ve landed the baseline Knip setup.

DMartens: Then I think we should pass the configuration to the `mocha` plugin:
```json
{
    // ... other knip properties
	"mocha": {
		"entry": ["tests/**/*.test.js"]
	}
}
```

Pixel998: I moved the test file glob into `mocha.entry`

DMartens (approved):
Changes LGTM, thanks (left open as a review from the TSC was requested).

Tanujkanti4441:
Hi @Pixel998, could you take a look at the conflicts?

package.json
```diff
@@ -74,19 +75,21 @@
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.18.2",
     "@eslint/json": "^1.1.0",
+    "@types/node": "^20.19.0",
```
fasttime, line 78: What is this needed for?

Pixel998: `@types/node` is a peer dependency of Knip.

PR #391, opened 2026-03-04
fix: allow CSS-wide keywords in font-family-fallbacks

### LINKED ISSUE #387: Bug: font-family-fallbacks reports an error for `font-family: inherit`
### Environment

ESLint version: v10.0.2
@eslint/css version:  0.14.1
Node version: v24.13
npm version: v11.9
Operating System: Debian stable

stylesheet

### What did you do?

<details>

```

export default [
    {
        files: ["**/*.css"],
        plugins: { css },
        language: "css/css",
        rules: {
            "css/font-family-fallbacks": "error",
        },
    },
];
```
</details>

```css
body {
    font-family: inherit;
    color: red;
}
```

No error, `inherit` shouldn't trigger the rule

### What actually happened?

Error:
```
  2:18  error  Use fallback fonts and a generic font last  css/font-family-fallbacks
```

https://github.com/julienw/eslint-csss-font-family-issue

### Participation

- [ ] I am willing to submit a pull request for this issue.

### AI acknowledgment

### Additional comments

_No response_

sethamus:
#### Prerequisites checklist

#### AI acknowledgment

This PR fixes false positives in `css/font-family-fallbacks` by allowing CSS-wide keywords (e.g. `inherit`, `initial`, `unset`, `revert`, `revert-layer`) for `font-family`/`font`, since they shouldn’t be required to include fallback/generic fonts.

- Updated the rule to treat CSS-wide keywords as always valid (including when they come from `var()` values).
- Added tests covering `font-family` and shorthand `font` CSS-wide keyword cases.
- Updated the rule docs to mention CSS-wide keywords are allowed.

#### Related Issues

Fixes #387

### DIFF
--- a/src/rules/font-family-fallbacks.js
+++ b/src/rules/font-family-fallbacks.js
@@ -33,30 +33,27 @@ const genericFonts = new Set([
 	"fangsong",
 ]);
 
-const cssWideKeywords = new Set([
-	"inherit",
-	"initial",
-	"unset",
-	"revert",
-	"revert-layer",
-]);
-
 /**
  * Check if the value is a CSS-wide keyword.
  * @param {string} value The value to check.
+ * @param {Set<string>} cssWideKeywords The CSS-wide keywords to check against.
  * @returns {boolean} True if the value is a CSS-wide keyword, false otherwise.
  */
-function isCSSWideKeyword(value) {
+function isCSSWideKeyword(value, cssWideKeywords) {
 	return cssWideKeywords.has(value.trim().toLowerCase());
 }
 
 /**
  * Check if the node is an identifier with a CSS-wide keyword.
  * @param {Object} node The node to check.
+ * @param {Set<string>} cssWideKeywords The CSS-wide keywords to check against.
  * @returns {boolean} True if the node is a CSS-wide keyword identifier, false otherwise.
  */
-function isCSSWideKeywordIdentifier(node) {
-	return node.type === "Identifier" && isCSSWideKeyword(node.name);
+function isCSSWideKeywordIdentifier(node, cssWideKeywords) {
+	return (
+		node.type === "Identifier" &&
+		isCSSWideKeyword(node.name, cssWideKeywords)
+	);
 }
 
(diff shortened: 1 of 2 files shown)


docs/rules/font-family-fallbacks.md
```diff
@@ -12,7 +12,7 @@ It is a best practice to use a fallback font and a generic font last in the `fon
 
 ## Rule Details
 
-This rule enforces the use of fallback fonts and a generic font last in `font-family` and `font` property.
+This rule enforces the use of fallback fonts and a generic font last in `font-family` and `font` property. Global values (`inherit`, `initial`, `revert`, `revert-layer`, `unset`) are always allowed.
```
Tanujkanti4441, line 15: Can you add some examples as well for this?

sethamus: Done!

src/rules/font-family-fallbacks.js
```diff
@@ -33,6 +33,32 @@ const genericFonts = new Set([
 	"fangsong",
 ]);
 
+const cssWideKeywords = new Set([
```
DMartens: Should we use `sourceCode.lexer.cssWideKeywordsSyntax.map` instead of this set?
This would allow users to fork the syntax and add css-wide keyword.

sethamus: What’s the motivating use case for extending `cssWideKeywords`?

nzakas: I think this is a good idea for maintenance purposes. It gives us one less spot to update should things change.

Pixel998: I think we should use `lexer.cssWideKeywords` here instead of `cssWideKeywordsSyntax`. This does currently cause a TypeScript error, so I’ve opened https://github.com/eslint/csstree/pull/128 to update the typings.

sethamus: I updated the rule to use `lexer.cssWideKeywords`.

PR #401, opened 2026-03-19
fix: correctly extract baseline support for CSS functions

xbinaryx:
#### Prerequisites checklist

#### AI acknowledgment

#### What did you do?

```css
a {
	background-image: image("foo.png", red);
}
```

`use-baseline` should report `image()` as a non-baseline CSS function.

#### What actually happened?

No lint error was reported for `image()`.

This PR fixes CSS function baseline extraction in the generator.

The previous logic would treat CSS data types as CSS functions when reading `css.types.*` compat keys. In practice, that meant `image()` could fail to report correctly because `css.types.image` (`<image>`) was being conflated with the `image()` function.

Refactored function extraction so CSS functions are handled explicitly, regenerated `baseline-data.js`, and added regression coverage.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/image/image


nzakas (approved):
LGTM. Would like @DMartens to verify before merging.

DMartens (approved):
Changes LGTM, thanks.

PR #445, opened 2026-05-07
feat: replace fix with suggestions in `no-duplicate-imports`

### LINKED ISSUE #441: Rule Change: Avoid fix for duplicate imports with condition

no-duplicate-imports

Implement suggestions

Other

### Example code

```css
@import 'test.css' print;
@import 'test.css';

/* OR */

@import 'test.css' (min-width: 800px);
@import 'test.css';

/* OR */

@import 'test.css' layer(base);
@import 'test.css' layer(components);
```

Right now, Rule reports the second import as a duplicate and removes it as an autofix.  But what if one @import has some import condition with it and the other doesn't, or both have different import conditions to apply? Then, in such conditions, autofix can impact the functionality.

The rule will not apply autofix when it finds duplicate imports with different import conditions.

### Participation

### AI acknowledgment

### Additional comments

_No response_

Tanujkanti4441:
#### Prerequisites checklist

#### AI acknowledgment

To remove the fix for duplicate imports that have conditions and apply suggestions to them.

Applied suggestions to the duplicate imports instead of fix.

```css
@import 'test.css' print;
@import 'test.css';
```
Now, for this code rule will propose two suggestions to remove anyone the duplicates with following messages:
For first one: `Remove duplicate @import rule with condition(s) - print.`
For second one: `Remove duplicate @import rule without conditions.`

```css
@import 'test.css' print;
@import 'test.css' print, screen;
```
Same for this code:
For first one: `Remove duplicate @import rule with condition(s) - print.`
For second one: `Remove duplicate @import rule with condition(s) - print, screen.`

```css
@import 'test.css' print;
@import 'test.css' print;
```
If both duplicates have the same conditions, then the rule will apply autofix and will remove the second one. For now, the autofix only works if the conditions in both then imports are in the same order.

Should we apply the autofix if imports have same conditions but in different order as well?

#### Related Issues
fixes #441

### DIFF
--- a/src/rules/no-duplicate-imports.js
+++ b/src/rules/no-duplicate-imports.js
@@ -57,6 +57,25 @@ function getImportConditions(importNode, sourceCode) {
 	return importConditions;
 }
 
+/**
+ * Get the fix for a duplicate import statement.
+ * @param {Object} fixer The fixer object.
+ * @param {string} text The full text of the source code.
+ * @param {number} start The start index of the import statement to fix.
+ * @param {number} end The end index of the import statement to fix.
+ * @param {boolean} condition A boolean indicating whether the import statement has conditions that differ from the original import.
+ * @returns {Object|null} A fix object if a fix is applicable, or null if no fix should be applied.
+ */
+function getFixForImport(fixer, text, start, end, condition) {
+	const removeEnd = getImportEnd(text, end);
+
+	if (condition) {
+		return fixer.removeRange([start, removeEnd]);
+	}
+
+	return null;
+}
+
 //-----------------------------------------------------------------------------
 // Rule
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/rules/no-duplicate-imports.js
DMartens: The [syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@import#syntax) of `@import` also allows specifying a layer and a supports condition (so not everything after is a condition).
Or should this be handled in a follow-up?

DMartens: We could make `imports` an array now instead of converting it from a set to an array for `Array#some/find`.

src/rules/no-duplicate-imports.js
```diff
@@ -95,54 +93,112 @@ export default {
...
+						firstImportConditions.every(
+							(condition, index) =>
+								condition === duplicateImportConditions[index],
+						);
+
 					context.report({
 						loc: node.loc,
 						messageId: "duplicateImport",
 						data: { url },
 						fix(fixer) {
```
DMartens, line 155: As the fixer is the same for the autofix and suggestion, I think we should extract this logic to a function.

src/rules/no-duplicate-imports.js
Tanujkanti4441: This PR already includes the layer and supports conditions. I have added tests regarding this also. So, I think we can just update the word `condition` to something more appropriate? like `modifier`?

src/rules/no-duplicate-imports.js
Tanujkanti4441: Done!

src/rules/no-duplicate-imports.js
DMartens: Sorry for the delay. I did not see those tests when writing this comment.
I agree to that changing the function name to `getImportModifiers` (I cannot think of something better than `modifier`).

Tanujkanti4441: Have changed word `condition` to `modifier` everywhere.

DMartens (approved):
Changes LGTM, thanks.

PR #466, opened 2026-05-30
fix: reporting font-size unit with percentage in line-height

Tanujkanti4441:
#### Prerequisites checklist

#### AI acknowledgment

In `relative-font-units` this PR fixes the reporting of font size with the correct unit when line-height is used with a percentage, such as:

```css
/* eslint css/relative-font-units: ["error", { allowUnits: ["rem"] }] */

a {
    font: 1rem/120% Arial, sans-serif;
 }
```
Right now, the rule is reporting this.
Added a condition to check whether the percentage is used with font-size or line-height.

### DIFF
--- a/src/rules/relative-font-units.js
+++ b/src/rules/relative-font-units.js
@@ -139,8 +139,22 @@ export default {
 							child => child.type === "Identifier",
 						);
 						const percentageNode = value.children.find(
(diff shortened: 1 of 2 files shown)


DMartens (approved):
Changes LGTM, thanks.

PR #471, opened 2026-06-12
fix: false negative when font used as identifier

Tanujkanti4441:
#### Prerequisites checklist

#### AI acknowledgment

To fix the false negative in `relative-font-units` when an identifier is used to set font size in `font` property, it fails to report the identifier font if there is another identifier used before it like `italic` or `bold` etc.

```css
"a { font: italic small Arial, sans-serif; }",
```

Right now, the rule is not reporting this.

Fixed the logic to find the disallowed identifier font size in `font`.


DMartens (approved):
Changes LGTM, thanks.

PR #474, opened 2026-06-19
feat: add suggestions to prefer-logical-properties

k-yle:
#### Prerequisites checklist

#### AI acknowledgment

- [ ] ~~(If the above is not checked) I have reviewed the AI-generated content before submitting.~~

IDE suggestions for the `prefer-logical-properties` rule

added a fixer to all 3 cases. 

#### Related Issues

none

### DIFF
--- a/src/rules/prefer-logical-properties.js
+++ b/src/rules/prefer-logical-properties.js
@@ -4,7 +4,7 @@
 
 /**
  * @import { CSSRuleDefinition } from "../types.js"
- * @typedef {"notLogicalProperty" | "notLogicalValue" | "notLogicalUnit"} PreferLogicalPropertiesMessageIds
+ * @typedef {"notLogicalProperty" | "notLogicalValue" | "notLogicalUnit" | "replaceWithLogicalProperty" | "replaceWithLogicalValue" | "replaceWithLogicalUnit"} PreferLogicalPropertiesMessageIds
  * @typedef {[{
  *     allowProperties?: string[],
  *     allowUnits?: string[]
@@ -137,7 +137,7 @@ export default /** @satisfies {PreferLogicalPropertiesRuleDefinition} */ ({
 	meta: {
 		type: "problem",
 
-		fixable: "code",
+		hasSuggestions: true,
 
 		docs: {
 			description: "Enforce the use of logical properties",
@@ -181,6 +181,12 @@ export default /** @satisfies {PreferLogicalPropertiesRuleDefinition} */ ({
 				"Expected logical value '{{replacement}}' instead of '{{value}}'.",
 			notLogicalUnit:
 				"Expected logical unit '{{replacement}}' instead of '{{unit}}'.",
+			replaceWithLogicalProperty:
+				"Replace '{{original}}' with logical property '{{replacement}}'.",
+			replaceWithLogicalValue:
+				"Replace '{{original}}' with logical value '{{replacement}}'.",
+			replaceWithLogicalUnit:
+				"Replace '{{original}}' with logical unit '{{replacement}}'.",
 		},
 	},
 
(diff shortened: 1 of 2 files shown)


DMartens:
Thank you for the PR.
In the future please open an issue before changing rules, so we can discuss the potential changes.
In this case the rule seems to should have a fixer (`meta.fixable` is set) but the fixer was not implemented. The corresponding stylelint rule [property-layout-mappings](https://stylelint.io/user-guide/rules/property-layout-mappings/) also provides an autofix.
Are there cases where an autofix would be unsafe?

k-yle:
an automatic fix is definitely not safe, it must be a manually reviewed suggestion.

Even if you design for RTL from day 1, there are always cases where you want something to be fixed to the the same side, regardless of LTR or RTL.

DMartens:
I agree, in the case the rule is enabled without the underlying mindset, it can cause breakages.
Can you please remove `meta.fixable` property, otherwise the changes LGTM.

k-yle:
sure, done

DMartens (dismissed):
Changes LGTM, thanks. Leaving open for a second review as it is a feature.

src/rules/prefer-logical-properties.js
```diff
@@ -181,6 +181,7 @@ export default /** @satisfies {PreferLogicalPropertiesRuleDefinition} */ ({
 				"Expected logical value '{{replacement}}' instead of '{{value}}'.",
 			notLogicalUnit:
 				"Expected logical unit '{{replacement}}' instead of '{{unit}}'.",
+			fixer: "Replace '{{original}}' with '{{replacement}}'.",
```
Pixel998: Since these suggestions are replacing different CSS concepts, I think separate IDs would read more clearly:

```js
replaceWithLogicalProperty:
	"Replace with logical property '{{replacement}}'.",
replaceWithLogicalValue:
	"Replace with logical value '{{replacement}}'.",
replaceWithLogicalUnit:
	"Replace with logical unit '{{replacement}}'.",

Comment (k-yle) [reply src/rules/prefer-logical-properties.js]: should we keep the original value in the message too? for example:

> "Replace '{{original}}' with the logical unit '{{replacement}}'."

Comment (Pixel998) [reply src/rules/prefer-logical-properties.js]: Sounds good.
```js
replaceWithLogicalProperty:
	"Replace '{{original}}' with logical property '{{replacement}}'.",
replaceWithLogicalValue:
	"Replace '{{original}}' with logical value '{{replacement}}'.",
replaceWithLogicalUnit:
	"Replace '{{original}}' with logical unit '{{replacement}}'.",
```

src/rules/prefer-logical-properties.js
```diff
@@ -4,7 +4,7 @@
 
 /**
  * @import { CSSRuleDefinition } from "../types.js"
- * @typedef {"notLogicalProperty" | "notLogicalValue" | "notLogicalUnit"} PreferLogicalPropertiesMessageIds
+ * @typedef {"notLogicalProperty" | "notLogicalValue" | "notLogicalUnit" | "replaceWithLogicalProperty" | "replaceWithLogicalValue" | "replaceWithLogicalUnit"} PreferLogicalPropertiesMessageIds
```
k-yle, line 7:
```suggestion
 * @typedef {`${"replaceWith" | "not"}Logical${"Property" | "Value" | "Unit"}`} PreferLogicalPropertiesMessageIds
```
this is more concise, but [`jsdoc/valid-types`](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/valid-types.md) erroenously claims that the syntax is invalid 🤷

src/rules/prefer-logical-properties.js
k-yle: cool, done

Pixel998 (approved):
LGTM, thanks! Leaving open for @DMartens.

DMartens (approved):
Changes LGTM, thank you for contributing.

Issue #477, opened 2026-06-24
Rule Change: Allow no-empty-blocks to ignore comment-only blocks


Pixel998:
### What rule do you want to change?

no-empty-blocks

### What change do you want to make?

Generate fewer warnings

### How do you think the change should be implemented?

A new option

### Example code

```css
@media (min-width: 1200px) {
  .sidebar { /* Fix this later */ }
}
```

### What does the rule currently do for this code?

The rule reports the block as empty.

The rule currently treats comments as not being content, so a block containing only a comment is reported. There is no option to change this behavior.

### What will the rule do after it's changed?

With a new option, users could choose for comment-only blocks to be treated as non-empty, so the example above would not be reported.

### Participation

- [x] I am willing to submit a pull request to implement this change.

### AI acknowledgment

- [x] I did not use AI to generate this issue report.
- [ ] (If the above is not checked) I have reviewed the AI-generated content before submitting.

### Additional comments

This would bring `no-empty-blocks` closer to Stylelint’s `block-no-empty`, which does not report comment-only blocks by default: https://stylelint.io/user-guide/rules/block-no-empty/

DMartens:
I agree with the change to add an option for this to be consistent with the respective stylelint rule.
For clarity should blocks with only a comment be allowed or disallowed by default?

Pixel998:
I think we should disallow comment-only blocks by default (keeping the current behavior).

Tanujkanti4441:
Idea sounds good to me. We can name it something like `allowCommentOnlyBlocks`?
And how about if the rule provides a suggestion as well to add a comment inside the empty block, only if this option is enabled?

nzakas:
For comparison, the ESLint [`no-empty`](https://eslint.org/docs/latest/rules/no-empty) rule has always allowed comment-only blocks by default. I'd vote for having that also be the default behavior here with an `ignoreCommentOnlyBlocks` option to turn that behavior off if desired.

Tanujkanti4441:
>For the purposes of this rule, comments do not count as content and this rule warns when the only content inside of a block is a comment.

I was skeptical about allowing comments by default because of this sentence in the rule docs.

But since it makes sense to me to make the rule's behavior similar `no-empty` core rule i think we can go with `ignoreCommentOnlyBlocks` option with default `true` and `no-empty` also provides the suggestion to add a comment inside blocks.

PR #479, opened 2026-06-27
feat: add suggestions to no-empty-blocks

### LINKED ISSUE #476: Rule Change: autofix for no-empty-blocks

no-empty-blocks

Implement a `fix` xor `suggestions`

A new default behavior

### Example code

happy to send a PR for this. as far as i know:
- an empty block is safe to delete. For example `a {}` -> ` `
- for [`@layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer)s, it's safe to remove the curly braces. For example: `@layer defaults {}` -> `@layer defaults`
- other blocks like `@media name {}` would have no autofix. 

if the block contains comments, it would be a `suggestion`, otherwise it would be a `fix`.

no autofix nor suggestions

has autofix and suggestions

### Participation

### AI acknowledgment

- [ ] ~~(If the above is not checked) I have reviewed the AI-generated content before submitting.~~

### Additional comments

&shy;-

k-yle:
#### Prerequisites checklist

#### AI acknowledgment

- [ ] ~~(If the above is not checked) I have reviewed the AI-generated content before submitting.~~

Add sugggestions to `no-empty-blocks`. There are two different cases:

* `@layer name {}`  becomes `@layer name`
* all other empty blocks like `a {}` become ` `. This included unnamed layers such as `@layer {}`. 

#### Related Issues

Closes #476

- We don't check if a block contains comments, because that would be convered by #477
- To the best of my knowledge, deleting all these different types of empty blocks is safe (except [`@layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer) as mentioned above). I have asked a senior colleague and Claude Opus 4.8, both of whom agreed.

### DIFF
--- a/src/rules/no-empty-blocks.js
+++ b/src/rules/no-empty-blocks.js
@@ -53,21 +53,42 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
 							? [
 									{
 										messageId: "convertToStatement",
-										fix: fixer =>
-											fixer.replaceTextRange(
+										fix(fixer) {
+											const commentsBeforeBlock =
+												context.sourceCode.comments.filter(
+													comment =>
+														comment.loc.start
+															.offset >=
+															parent.prelude.loc
+																.end.offset &&
+														comment.loc.end
+															.offset <=
+															node.loc.start
+																.offset,
+												);
+											const lastComment =
+												commentsBeforeBlock.at(-1);
+
+											return fixer.replaceTextRange(
 												[
-													parent.prelude.loc.end
-														.offset,
+													lastComment
+														? lastComment.loc.end
+																.offset
+														: parent.prelude.loc.end
+																.offset,
 													node.loc.end.offset,
 												],
 												";",
-											),
+											);
+										},
 									},
 								]
 							: [
 									{
 										messageId: "removeRule",
-										fix: fixer => fixer.remove(parent),
+										fix(fixer) {
+											return fixer.remove(parent);
+										},
 									},
 								],
 					});
--- a/src/rules/no-empty-blocks.js
+++ b/src/rules/no-empty-blocks.js
@@ -9,7 +9,7 @@
 
 /**
  * @import { CSSRuleDefinition } from "../types.js"
- * @typedef {"emptyBlock"} NoEmptyBlocksMessageIds
+ * @typedef {"emptyBlock" | "removeRule" | "convertToStatement"} NoEmptyBlocksMessageIds
  * @typedef {CSSRuleDefinition<{ RuleOptions: [], MessageIds: NoEmptyBlocksMessageIds }>} NoEmptyBlocksRuleDefinition
  */
 
@@ -21,6 +21,8 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
 	meta: {
 		type: "problem",
 
+		hasSuggestions: true,
+
 		docs: {
 			description: "Disallow empty blocks",
 			recommended: true,
@@ -29,6 +31,8 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
 
 		messages: {
 			emptyBlock: "Unexpected empty block found.",
+			removeRule: "Remove the empty rule.",
+			convertToStatement: "Convert to layer statement.",
 		},
 	},
 
@@ -36,9 +40,36 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
 		return {
 			Block(node) {
 				if (node.children.length === 0) {
+					const parent = context.sourceCode.getParent(node);
+					const isNamedAtLayer =
+						parent.type === "Atrule" &&
+						parent.name.toLowerCase() === "layer" &&
+						parent.prelude;
+
 					context.report({
 						loc: node.loc,
 						messageId: "emptyBlock",
+						suggest: isNamedAtLayer
+							? [
+									{
+										messageId: "convertToStatement",
+										fix: fixer =>
+											fixer.replaceTextRange(
+												[
+													parent.prelude.loc.end
+														.offset,
+													node.loc.end.offset,
+												],
+												";",
+											),
+									},
+								]
+							: [
+									{
+										messageId: "removeRule",
+										fix: fixer => fixer.remove(parent),
+									},
+								],
 					});
 				}
 			},
--- a/tests/rules/no-empty-blocks.test.js
+++ b/tests/rules/no-empty-blocks.test.js
@@ -163,10 +163,68 @@ ruleTester.run("no-empty-blocks", rule, {
 			],
 		},
 		{
+			code: "@layer defaults /* comment */ { }",
+			errors: [
+				{
+					messageId: "emptyBlock",
+					line: 1,
+					column: 31,
+					endLine: 1,
+					endColumn: 34,
+					suggestions: [
+						{
+							messageId: "convertToStatement",
+							output: "@layer defaults /* comment */;",
+						},
+					],
+				},
+			],
+		},
+		{
+			code: "@layer defaults /*\n * comment\n */ { }",
+			errors: [
+				{
+					messageId: "emptyBlock",
+					line: 3,
+					column: 5,
+					endLine: 3,
+					endColumn: 8,
+					suggestions: [
+						{
+							messageId: "convertToStatement",
+							output: "@layer defaults /*\n * comment\n */;",
+						},
+					],
+				},
+			],
+		},
+		{
+			code: "@layer defaults /* first */ /* second */ { }",
+			errors: [
+				{
+					messageId: "emptyBlock",
+					line: 1,
+					column: 42,
+					endLine: 1,
+					endColumn: 45,
+					suggestions: [
+						{
+							messageId: "convertToStatement",
+							output: "@layer defaults /* first */ /* second */;",
+						},
+					],
+				},
+			],
+		},
+		{
 			code: "@LAYER foo{}",
 			errors: [
 				{
 					messageId: "emptyBlock",
+					line: 1,
+					column: 11,
+					endLine: 1,
+					endColumn: 13,
 					suggestions: [
 						{
 							messageId: "convertToStatement",
@@ -181,6 +239,10 @@ ruleTester.run("no-empty-blocks", rule, {
 			errors: [
 				{
 					messageId: "emptyBlock",
+					line: 2,
+					column: 2,
+					endLine: 2,
+					endColumn: 5,
 					suggestions: [
 						{
 							messageId: "convertToStatement",
--- a/tests/rules/no-empty-blocks.test.js
+++ b/tests/rules/no-empty-blocks.test.js
@@ -34,6 +34,7 @@ ruleTester.run("no-empty-blocks", rule, {
 					column: 3,
 					endLine: 1,
 					endColumn: 6,
+					suggestions: [{ messageId: "removeRule", output: "" }],
 				},
 			],
 		},
@@ -46,6 +47,7 @@ ruleTester.run("no-empty-blocks", rule, {
 					column: 3,
 					endLine: 1,
 					endColumn: 20,
+					suggestions: [{ messageId: "removeRule", output: "" }],
 				},
 			],
 		},
@@ -58,6 +60,7 @@ ruleTester.run("no-empty-blocks", rule, {
 					column: 3,
 					endLine: 2,
 					endColumn: 2,
+					suggestions: [{ messageId: "removeRule", output: "" }],
 				},
 			],
 		},
@@ -70,6 +73,7 @@ ruleTester.run("no-empty-blocks", rule, {
 					column: 3,
 					endLine: 2,
 					endColumn: 3,
+					suggestions: [{ messageId: "removeRule", output: "" }],
 				},
 			],
 		},
@@ -82,6 +86,12 @@ ruleTester.run("no-empty-blocks", rule, {
 					column: 14,
 					endLine: 1,
 					endColumn: 17,
+					suggestions: [
+						{
+							messageId: "removeRule",
+							output: "",
+						},
+					],
 				},
 			],
 		},
@@ -94,6 +104,12 @@ ruleTester.run("no-empty-blocks", rule, {
 					column: 3,
 					endLine: 1,
 					endColumn: 6,
+					suggestions: [
+						{
+							messageId: "removeRule",
+							output: "\n@media print { \nb { } \n}", // only a{} was removed
+						},
+					],
 				},
 				{
 					messageId: "emptyBlock",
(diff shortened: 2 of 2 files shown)


src/rules/no-empty-blocks.js
```diff
@@ -39,6 +42,25 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
...
+										.at(-1);
+									const isAtLayer =
+										parent.type === "Atrule" &&
+										parent.name === "layer";
+
+									const nodeToReplace = isAtLayer
+										? node
+										: parent;
+
+									return fixer.replaceText(nodeToReplace, "");
```
DMartens:
```suggestion
									return fixer.remove(nodeToReplace);
```
There is a text fixer for removing which make this more semantic.
Could you please also rename `nodeToReplace` to `nodeToRemove`?

src/rules/no-empty-blocks.js
```diff
@@ -39,6 +42,25 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
 					context.report({
 						loc: node.loc,
 						messageId: "emptyBlock",
+						suggest: [
+							{
+								messageId: "deleteEmptyBlock",
+								fix(fixer) {
+									const parent = context.sourceCode
```
DMartens: We should use `context.sourceCode.getParent(node)` here instead of `getAncestors`.

DMartens:
Thank you for the PR. Just found two small refactorings using more appropriate methods.
Also checked and tried ways to create a parse error / changing AST after removal but could not find one.
Would like the second reviewer to focus on this.

src/rules/no-empty-blocks.js
k-yle: thanks, both suggestions applied

src/rules/no-empty-blocks.js
```diff
@@ -29,6 +31,7 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
 
 		messages: {
 			emptyBlock: "Unexpected empty block found.",
+			deleteEmptyBlock: "Delete the empty block.",
```
Pixel998: This is slightly misleading since `fixer.remove(parent)` removes the entire rule, not just the block.

src/rules/no-empty-blocks.js
```diff
@@ -39,6 +42,22 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
...
+								messageId: "deleteEmptyBlock",
+								fix(fixer) {
+									const parent =
+										context.sourceCode.getParent(node);
+									const isAtLayer =
+										parent.type === "Atrule" &&
+										parent.name === "layer";
+
+									return fixer.remove(
+										isAtLayer ? node : parent,
```
Pixel998: Removing the block of a named layer can cause a parsing error ([explorer](https://explorer.eslint.org/#eslint-explorer=v2.eyJzdGF0ZSI6eyJ0b29sIjoiYXN0IiwiY29kZSI6eyJqYXZhc2NyaXB0IjoiLyoqXG4gKiBUeXBlIG9yIHBhc3RlIHNvbWUgSmF2YVNjcmlwdCBoZXJlIHRvIGxlYXJuIG1vcmUgYWJvdXRcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXG4gKiBcbiAqIFRoZSB0aHJlZSB0YWJzIGFyZTpcbiAqIFxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbiAqICAgY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0LlxuICogLSBTY29wZSAtIFRoZSBzY29wZSBzdHJ1Y3R1cmUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhbiBiZVxuICogICB1c2VmdWwgdG8gdW5kZXJzdGFuZCBob3cgdmFyaWFibGVzIGFyZSBkZWZpbmVkIGFuZFxuICogICB3aGVyZSB0aGV5IGFyZSB1c2VkLlxuICogLSBDb2RlIFBhdGggLSBUaGUgY29kZSBwYXRoIHN0cnVjdHVyZSBvZiB0aGUgY29kZSwgd2hpY2hcbiAqICAgY2FuIGJlIHVzZWZ1bCB0byB1bmRlcnN0YW5kIGhvdyB0aGUgY29kZSBpcyBleGVjdXRlZC5cbiAqIFxuICogWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBKYXZhU2NyaXB0IGNvZGUgaXMgaW50ZXJwcmV0ZWRcbiAqIGJ5IGNsaWNraW5nIFwiSmF2YVNjcmlwdFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcbiAqIG9wdGlvbnMuXG4gKi9cblxuaW1wb3J0IGpzIGZyb20gXCJAZXNsaW50L2pzXCI7XG5pbXBvcnQgeyBkZWZpbmVDb25maWcgfSBmcm9tIFwiZXNsaW50L2NvbmZpZ1wiO1xuXG5mdW5jdGlvbiBnZXRDb25maWcoKSB7XG4gICAgcmV0dXJuIHtcbiAgICAgICAgcnVsZXM6IHtcbiAgICAgICAgICAgIFwicHJlZmVyLWNvbnN0XCI6IFwid2FyblwiXG4gICAgICAgIH1cbiAgICB9O1xufVxuXG5leHBvcnQgZGVmYXVsdCBkZWZpbmVDb25maWcoW1xuICAgIGpzLmNvbmZpZ3MucmVjb21tZW5kZWQsXG4gICAgZ2V0Q29uZmlnKClcbl0pOyIsImpzb24iOiIvKipcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxuICogdGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cbiAqXG4gKiBUaGUgdGFicyBhcmU6XG4gKlxuICogLSBBU1QgLSBUaGUgQWJzdHJhY3QgU3ludGF4IFRyZWUgb2YgdGhlIGNvZGUsIHdoaWNoIGNhblxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbiAqICAgY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0LlxuICpcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXG4gKiBieSBjbGlja2luZyBcIkpTT05cIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XG4gKiBvcHRpb25zLlxuICpcbiAqIFRoaXMgZXhhbXBsZSBpcyBpbiBKU09OQyBtb2RlLCB3aGljaCBhbGxvd3MgY29tbWVudHMuXG4gKi9cblxue1xuICAgIFwia2V5MVwiOiBbdHJ1ZSwgZmFsc2UsIG51bGxdLFxuICAgIFwia2V5MlwiOiB7XG4gICAgICAgIFwia2V5M1wiOiBbMSwgMiwgXCIzXCIsIDFlMTAsIDFlLTNdXG4gICAgfVxufSIsIm1hcmtkb3duIjoiPCEtLVxuVHlwZSBvciBwYXN0ZSBzb21lIE1hcmtkb3duIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxudGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cblxuVGhlIHRhYnMgYXJlOlxuXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbmNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cblxuWW91IGNhbiBjaGFuZ2UgdGhlIHdheSB0aGF0IHRoZSBNYXJrZG93biBjb2RlIGlzIGludGVycHJldGVkXG5ieSBjbGlja2luZyBcIk1hcmtkb3duXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxub3B0aW9ucy5cblxuVGhpcyBleGFtcGxlIGlzIGluIENvbW1vbk1hcmsgbW9kZS5cbi0tPlxuXG4jIEVTTGludCBNYXJrZG93biBFeGFtcGxlXG5cblRoaXMgaXMgYW4gZXhhbXBsZSBvZiBhIE1hcmtkb3duIGZpbGUgdGhhdCBjYW4gYmUgcGFyc2VkXG5ieSBFU0xpbnQuIE1hcmtkb3duIGlzIGEgc2ltcGxlIG1hcmt1cCBsYW5ndWFnZSB0aGF0IGlzXG5vZnRlbiB1c2VkIGZvciBkb2N1bWVudGF0aW9uLlxuXG4jIyBGZWF0dXJlc1xuXG4tIE1ha2UgdGhpbmdzICppdGFsaWMqLCAqKmJvbGQqKiwgb3IgYGNvZGVgXG4tIENyZWF0ZSBbbGlua3NdKGh0dHBzOi8vZXNsaW50Lm9yZylcbi0gU3VwcG9ydHMgSFRNTCA8c3BhbiBzdHlsZT1cImNvbG9yOiByZWQ7XCI-ZWxlbWVudHM8L3NwYW4-XG4tIExpc3RzXG4gIC0gTmVzdGVkIGxpc3RzIiwiY3NzIjoiQGxheWVyIGZvb1xuXG5hIHtcbiAgY29sb3I6IHJlZDtcbn1cbiIsImh0bWwiOiI8IURPQ1RZUEUgaHRtbD5cbjwhLS1cblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxudGhlIHN0YXRpYyBhbmFseXNpcyB0aGF0IEVTTGludCBjYW4gZG8gZm9yIHlvdS5cblxuVGhlIHRhYnMgYXJlOlxuXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcbmNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cbi0tPlxuXG48aHRtbCBsYW5nPVwiZW5cIj5cbiAgICA8aGVhZD5cbiAgICAgICAgPG1ldGEgY2hhcnNldD1cIlVURi04XCI-XG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cbiAgICA8L2hlYWQ-XG4gICAgPGJvZHk-XG4gICAgICAgIDxwPlRleHQ8L3A-XG4gICAgPC9ib2R5PlxuPC9odG1sPiJ9LCJsYW5ndWFnZSI6ImNzcyIsImpzT3B0aW9ucyI6eyJwYXJzZXIiOiJlc3ByZWUiLCJzb3VyY2VUeXBlIjoibW9kdWxlIiwiZXNWZXJzaW9uIjoibGF0ZXN0IiwiaXNKU1giOnRydWV9LCJqc29uT3B0aW9ucyI6eyJqc29uTW9kZSI6Impzb25jIiwiYWxsb3dUcmFpbGluZ0NvbW1hcyI6ZmFsc2V9LCJjc3NPcHRpb25zIjp7ImNzc01vZGUiOiJjc3MiLCJ0b2xlcmFudCI6ZmFsc2V9LCJtYXJrZG93bk9wdGlvbnMiOnsibWFya2Rvd25Nb2RlIjoiY29tbW9ubWFyayIsIm1hcmtkb3duRnJvbnRtYXR0ZXIiOiJvZmYiLCJtYXJrZG93bk1hdGgiOmZhbHNlfSwiaHRtbE9wdGlvbnMiOnsidGVtcGxhdGVFbmdpbmVTeW50YXgiOiJub25lIiwiZnJvbnRtYXR0ZXIiOmZhbHNlfSwid3JhcCI6dHJ1ZSwidmlld01vZGVzIjp7ImFzdFZpZXciOiJ0cmVlIiwic2NvcGVWaWV3IjoiZmxhdCIsInBhdGhWaWV3IjoiZ3JhcGgifSwicGF0aEluZGV4Ijp7ImluZGV4IjowLCJpbmRleGVzIjoxfSwiZXNxdWVyeVNlbGVjdG9yIjp7InNlbGVjdG9yIjoiIn19LCJ2ZXJzaW9uIjowfQ)). Instead, it needs to be replaced with a semicolon to become a valid layer statement. For unnamed layers, I think it is fine to remove the entire rule.

Because of this distinction, we should probably split this into two separate suggestions:

```js
removeRule: "Remove the empty rule.",
convertToStatement: "Convert to layer statement.",
```

Pixel998 (changes requested):

src/rules/no-empty-blocks.js
k-yle: thanks, both suggestions applied, and PR description updated

src/rules/no-empty-blocks.js
```diff
@@ -29,16 +31,38 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
...
 	},
 
 	create(context) {
 		return {
 			Block(node) {
 				if (node.children.length === 0) {
+					const parent = context.sourceCode.getParent(node);
+					const isNamedAtLayer =
+						parent.type === "Atrule" &&
+						parent.name === "layer" &&
```
Pixel998:
```suggestion
						parent.name.toLowerCase() === "layer" &&
```

At-rules are case-insensitive, so this will miss `@LAYER` or `@Layer`.

src/rules/no-empty-blocks.js
```diff
@@ -29,16 +31,38 @@ export default /** @satisfies {NoEmptyBlocksRuleDefinition} */ ({
...
+
 					context.report({
 						loc: node.loc,
 						messageId: "emptyBlock",
+						suggest: isNamedAtLayer
+							? [
+									{
+										messageId: "convertToStatement",
+										fix: fixer =>
+											fixer.replaceText(node, ";"),
```
Pixel998: The fixer currently only replaces the block node, which preserves any whitespace before `{}` and can produce output like `@layer foo ;`. Could we include the preceding whitespace in the replacement range so the fix produces `@layer foo;`?

Pixel998 (dismissed):

src/rules/no-empty-blocks.js
k-yle: sorry, i did indeed lose track of this PR. 

I did this using `node.prelude.loc.end`...`node.loc.end`, not sure if there's a more idiomatic solution

Pixel998: One edge case is that the replacement range also removes comments between the layer prelude and `{`, such as `@layer foo /* comment */ {}`. We need to preserve those comments, so I’ll push an update for that.

DMartens (approved):
Changes LGTM, thanks again. Leaving open for @Pixel998.

Pixel998:
LGTM, thanks!

PR #480, opened 2026-06-27
refactor: simplify prefer-logical-properties lookups

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

This PR refactors the `prefer-logical-properties` rule to reduce redundant map lookups.

- Reused local replacement variables for properties, values, and units instead of repeating map lookups.
- Updated suggestion message templates to use specific data keys: `property`, `value`, and `unit`.
- Updated rule tests to assert suggestion `messageId` and `data` instead of descriptions.


DMartens (approved):
Changes LGTM, thanks.

PR #507, opened 2026-07-20
fix: distinguish named timeline range keyframe selectors

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

#### What did you do?

```css
/* eslint css/no-duplicate-keyframe-selectors: "error" */

@keyframes in-and-out {
  entry 0% {}
  entry 100% {}
  exit 0% {}
  exit 100% {}
}
```

No duplicate-selector errors, because each range/percentage pair is distinct.

#### What actually happened?

The rule used only the selector’s first component (`entry` or `exit`), so it incorrectly reported `entry 100%` and `exit 100%` as duplicates.

This PR fixes false duplicate-selector reports for `@keyframes` selectors that include a named timeline range.

Updated `no-duplicate-keyframe-selectors` to use the complete keyframe selector when detecting duplicates.

### DIFF
--- a/tests/rules/no-duplicate-keyframe-selectors.test.js
+++ b/tests/rules/no-duplicate-keyframe-selectors.test.js
@@ -418,5 +418,53 @@ ruleTester.run("no-duplicate-keyframe-selectors", rule, {
 				},
 			],
 		},
+		{
+			code: dedent`@keyframes test {
+                entry 0% { opacity: 0; }
+                entry  0% { opacity: 1; }
+            }`,
+			errors: [
+				{
+					messageId: "duplicateKeyframeSelector",
+					data: { selector: "entry 0%" },
+					line: 3,
+					column: 5,
+					endLine: 3,
+					endColumn: 14,
+				},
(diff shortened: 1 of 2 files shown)


tests/rules/no-duplicate-keyframe-selectors.test.js
```diff
@@ -396,5 +402,21 @@ ruleTester.run("no-duplicate-keyframe-selectors", rule, {
 				},
 			],
 		},
+		{
```
DMartens, line 405: Can you please add more test cases for whitespace in between, e.g. "entry   0%" (otherwise we could just use `sourceCode.getText(selector)` instead of the stringify logic)

Pixel998: Added in https://github.com/eslint/css/pull/507/commits/fdc4687b2885aea42551a0fab52622b4728a523f.

DMartens (approved):
Changes LGTM, thanks.

PR #510, opened 2026-07-20
fix: skip validating values containing env() in no-invalid-properties

### LINKED ISSUE #509: Bug: no-invalid-properties reports "Unknown property" when the value contains env()
### Environment

ESLint version: v10.7.0
@eslint/css version: v1.4.0
Node version: v22.15.0
npm version: v11.12.1
Operating System: Windows 11

stylesheet

### What did you do?

<details>

```js

export default [
	{
		files: ["**/*.css"],
		plugins: { css },
		language: "css/css",
		rules: {
			"css/no-invalid-properties": "error",
		},
	},
];
```
</details>

```css
.a {
	padding-top: env(safe-area-inset-top);
}

.b {
	padding-top: calc(env(safe-area-inset-top) + 10px);
}
```

No errors. These are all valid declarations — env() is standard CSS and padding-top/width are obviously real properties.

### What actually happened?

Both declarations are reported as an unknown property:

```
  2:2  error  Unknown property 'padding-top' found  css/no-invalid-properties
  6:2  error  Unknown property 'padding-top' found  css/no-invalid-properties
```

css-tree's lexer throws a plain Error ("Matching for a tree with env() is not supported") when the value contains env(), and the error handling in no-invalid-properties treats any non-SyntaxMatchError as an unknown property when allowUnknownVariables is false. So the "env() not supported" error ends up reported as "Unknown property".

Setting allowUnknownVariables: true works around it, but the default hits pretty common CSS (safe-area-inset padding) and the message is wrong either way. Maybe values containing env() should just skip validation, like
unresolved var() does?

https://stackblitz.com/edit/stackblitz-starters-2hdjwmbx

### Participation

### AI acknowledgment

### Additional comments

_No response_

minseonkkim:
#### Prerequisites checklist

#### AI acknowledgment

Fixes a false positive in `no-invalid-properties`: any declaration whose value contains `env()` was reported as `Unknown property '...' found`, even for obviously valid properties like `padding-top`.

As noted in #509, the csstree fork bails out with a plain Error ("Matching for a tree with env() is not supported") when a value contains `env()`, but the rule was routing that error to the `unknownProperty` report.

I added an `isEnvMatchError()` helper to `src/util.js` and made the rule skip validation when that error occurs, since `env()` values come from the user agent and can't be validated anyway.

Unknown property names are still reported for declarations with `env()` values, because the lexer checks the property reference before the `env()` check — added an invalid test (`paddin-top: env(...)`) to lock that in, plus valid tests for `env()` alone, with a fallback, inside `calc()`, and in a shorthand.

#### Related Issues

fixes #509

`isEnvMatchError()` matches the lexer's error by its exact message string. Let me know if there's a better way to detect this error.

### DIFF
--- a/src/rules/no-invalid-properties.js
+++ b/src/rules/no-invalid-properties.js
@@ -7,7 +7,11 @@
 // Imports
 //-----------------------------------------------------------------------------
 
-import { isSyntaxMatchError, isSyntaxReferenceError } from "../util.js";
+import {
+	isSyntaxMatchError,
+	isSyntaxReferenceError,
+	isEnvMatchError,
+} from "../util.js";
 
 //-----------------------------------------------------------------------------
 // Type Definitions
@@ -463,7 +467,13 @@ export default /** @satisfies {NoInvalidPropertiesRuleDefinition} */ ({
 						});
 						return;
 					}
-
+					if (isEnvMatchError(error)) {
+						/*
+						 * env() values are provided by the user agent and
+						 * cannot be validated, so skip validation entirely.
+						 */
+						return;
+					}
 					if (
 						!allowUnknownVariables ||
 						isSyntaxReferenceError(error)
--- a/src/util.js
+++ b/src/util.js
@@ -32,3 +32,13 @@ export function isSyntaxMatchError(error) {
 export function isSyntaxReferenceError(error) {
 	return typeof error.reference === "string";
 }
+
+/**
+ * Determines if an error is the lexer error thrown when a value containing
+ * `env()` cannot be matched against a syntax definition.
+ * @param {Object} error The error object to check.
+ * @returns {boolean} True if the error is the `env()` match error, false if not.
+ */
+export function isEnvMatchError(error) {
+	return error.message === "Matching for a tree with env() is not supported";
+}
--- a/tests/rules/no-invalid-properties.test.js
+++ b/tests/rules/no-invalid-properties.test.js
@@ -30,6 +30,7 @@ ruleTester.run("no-invalid-properties", rule, {
 		"a { padding-top: calc(env(safe-area-inset-top) + 10px); }",
 		"a { width: env(titlebar-area-width); }",
 		"a { padding: env(safe-area-inset-top) 0 env(safe-area-inset-bottom) 0; }",
+		"a { padding: env(safe-area-inset-top) red }",
 		"a { color: red; background-color: blue; }",
 		"a { color: red; transition: none; }",
 		"body { --custom-property: red; }",
--- a/tests/rules/no-invalid-properties.test.js
+++ b/tests/rules/no-invalid-properties.test.js
@@ -25,6 +25,11 @@ const ruleTester = new RuleTester({
 ruleTester.run("no-invalid-properties", rule, {
 	valid: [
 		"a { color: red; }",
+		"a { padding-top: env(safe-area-inset-top); }",
+		"a { padding-top: env(safe-area-inset-top, 20px); }",
+		"a { padding-top: calc(env(safe-area-inset-top) + 10px); }",
+		"a { width: env(titlebar-area-width); }",
+		"a { padding: env(safe-area-inset-top) 0 env(safe-area-inset-bottom) 0; }",
 		"a { color: red; background-color: blue; }",
 		"a { color: red; transition: none; }",
 		"body { --custom-property: red; }",
@@ -276,6 +281,19 @@ ruleTester.run("no-invalid-properties", rule, {
 		"main { p:first-of-type, span { color: red; } }",
 	],
 	invalid: [
+		{
+			code: "a { paddin-top: env(safe-area-inset-top); }",
+			errors: [
+				{
+					messageId: "unknownProperty",
+					data: { property: "paddin-top" },
+					line: 1,
+					column: 5,
+					endLine: 1,
+					endColumn: 15,
+				},
+			],
+		},
 		{
 			code: "a { color: bar }",
 			errors: [
--- a/tests/rules/no-invalid-properties.test.js
+++ b/tests/rules/no-invalid-properties.test.js
@@ -25,12 +25,6 @@ const ruleTester = new RuleTester({
 ruleTester.run("no-invalid-properties", rule, {
 	valid: [
 		"a { color: red; }",
-		"a { padding-top: env(safe-area-inset-top); }",
-		"a { padding-top: env(safe-area-inset-top, 20px); }",
-		"a { padding-top: calc(env(safe-area-inset-top) + 10px); }",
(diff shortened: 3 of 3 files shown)


tests/rules/no-invalid-properties.test.js
```diff
@@ -25,6 +25,11 @@ const ruleTester = new RuleTester({
 ruleTester.run("no-invalid-properties", rule, {
 	valid: [
 		"a { color: red; }",
+		"a { padding-top: env(safe-area-inset-top); }",
+		"a { padding-top: env(safe-area-inset-top, 20px); }",
+		"a { padding-top: calc(env(safe-area-inset-top) + 10px); }",
+		"a { width: env(titlebar-area-width); }",
+		"a { padding: env(safe-area-inset-top) 0 env(safe-area-inset-bottom) 0; }",
```
DMartens: Could please add a valid test case for a declaration with multiple values but one of them is invalid (e.g. `a { padding: env(safe-area-inset-top) red }`) as we do not support partial validation.

minseonkkim: Thanks, good catch! Added it. I confirmed it stays valid since env() causes the whole value to skip validation.

src/util.js
```diff
@@ -32,3 +32,13 @@ export function isSyntaxMatchError(error) {
...
 }
+
+/**
+ * Determines if an error is the lexer error thrown when a value containing
+ * `env()` cannot be matched against a syntax definition.
+ * @param {Object} error The error object to check.
+ * @returns {boolean} True if the error is the `env()` match error, false if not.
+ */
+export function isEnvMatchError(error) {
+	return error.message === "Matching for a tree with env() is not supported";
```
lumirlumir, line 43: Non-blocking comment: relying on an internal error message seems fragile because changes in `csstree` could break this check. Error messages are generally not a stable API and may vary between versions.

Currently, there does not appear to be a reliable way to distinguish this unsupported `env()` matching case from other generic errors, so checking the message may be the most practical option for now. 

In the longer term, introducing a custom error class -- similar to the existing [`SyntaxReferenceError` and `SyntaxMatchError`](https://github.com/eslint/csstree/blob/bcf38310321086227f308a16d5295a1953ad9e7b/lib/lexer/error.js) -- or exposing a stable error code such as `ERR_LEXER_ENV_MATCH_UNSUPPORTED` would provide a more robust solution.

src/rules/no-invalid-properties.js
lumirlumir, line 367: Could we detect `env()` before processing nested `var()` functions? 

With the default options, `a { padding: env(safe-area-inset-top, var(--external-padding)); }` still reports `unknownVar`, because the nested `var()` is processed before `isEnvMatchError()` is reached. 

Since the fallback may never be selected, this remains a false positive and means that validation is not entirely skipped for values containing `env()`.

For example, if I add the following test case to `valid`:

```js
"a { padding: env(safe-area-inset-top, var(--external-padding)); }",
```

The following error occurs:

```sh
AssertionError [ERR_ASSERTION]: Should have no errors but had 1: [
  {
    ruleId: 'rule-to-test/no-invalid-properties',
    severity: 1,
    message: "Can't validate with unknown variable '--external-padding'.",
    line: 1,
    column: 43,
    messageId: 'unknownVar',
    endLine: 1,
    endColumn: 61
  }
]
```

lumirlumir (changes requested):

src/rules/no-invalid-properties.js
DMartens: Personally I think reporting the `var` is correct as this rule currently has two responsibilities: validating the syntax and checking whether custom properties are resolvable (In other words, there is currently not a rule only for checking if a custom property is defined locally).
The rule currently does not but hopefully does in the future support checking parts of the value (e.g. report "red" in `padding: env(...) red`)).

src/util.js
DMartens: This is an error only our [fork](https://github.com/eslint/csstree/blob/bcf38310321086227f308a16d5295a1953ad9e7b/lib/lexer/Lexer.js#L93) of csstree throws ([reference for upstream](https://github.com/csstree/csstree/blob/88e3d965c0b1628642a30a841745b410d6835052/lib/lexer/Lexer.js#L79)), so we could add an error code to this, but we can do this after this PR.

src/rules/no-invalid-properties.js
minseonkkim: Thanks for pointing this out! I agree this is currently a false positive if the fallback is never selected.

However, based on DMartens' comment, the current rule intentionally has two responsibilities: validating the syntax and checking whether custom properties are resolvable. Skipping nested var() validation inside env() would change that behavior.

I think it makes sense to revisit this once the rule supports partial validation of values, as mentioned.

lumirlumir: Okay, if so, can we add the following test case (which was mentioned above) to the `invalid` section to ensure that this is the intended behavior?

```js
"a { padding: env(safe-area-inset-top, var(--external-padding)); }",
```

Also, adding a comment like “Reporting `var()` inside `env()` here is intentional, but we can revisit it once this rule supports partial validation of values,” or something similar, seems helpful here.

minseonkkim: Added it to the invalid section along with a comment explaining why this is intentional. Thanks for the suggestion.

src/util.js
lumirlumir: @minseonkkim 

If you’d like to open an issue about it, that would be welcome.

lumirlumir (dismissed):
LGTM, thanks!

Would like @DMartens to verify before merging.

tests/rules/no-invalid-properties.test.js
```diff
@@ -276,6 +282,39 @@ ruleTester.run("no-invalid-properties", rule, {
 		"main { p:first-of-type, span { color: red; } }",
 	],
 	invalid: [
+		{
```
DMartens: These test cases should not be the first test cases as they check for specific scenarios.

minseonkkim: Thanks for the review! I've moved the env() test cases (including the new var()-in-env() case) after the existing generic cases in both arrays, grouped with a comment like the other scenario-specific test groups.

lumirlumir (approved):

DMartens (approved):
Changes LGTM, thanks.

src/util.js
minseonkkim: Opened eslint/csstree#141 for this. Thanks for the suggestion!

lumirlumir:
Disclosure for clarity: The author is a participant in the [open source contribution program OSSCA](https://github.com/eslint-ossca), which I lead.

The related issue and PR were both identified, written, and prepared by the participant.

PR #512, opened 2026-07-24
fix: detect equivalent selectors in `no-duplicate-keyframe-selectors`

Pixel998:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

#### AI acknowledgment

- [x] I did _not_ use AI to generate this PR.
- [ ] (If the above is not checked) I have reviewed the AI-generated content before submitting.

#### What is the purpose of this pull request?

This PR makes `no-duplicate-keyframe-selectors` report equivalent `from`/`0%` and `to`/`100%` keyframe selectors as duplicates.

#### What changes did you make? (Give an overview)

- Canonicalized `from` to `0%` and `to` to `100%` for duplicate detection.
- Updated the rule documentation.
- Added tests.

#### Related Issues

Fixes #508

PR #513, opened 2026-07-24
refactor: no-duplicate-keyframe-selectors use Set instead of Map

DMartens:
#### Prerequisites checklist

#### AI acknowledgment

While reviewing #512 I saw the unneeded `Map` usage.
As the value of the key is unused, it is more idiomatic to use a `Set`.

Use a `Map` instead of `Set` to keep track of duplicates.


lumirlumir (approved):
LGTM, thanks!

I’m marking the PR prefix as `refactor` since this is a production code change.

PR #517, opened 2026-07-28
feat: make font-family-fallbacks generic fonts case-insensitive

### LINKED ISSUE #516: Bug: `font-family-fallbacks` reports an error for generic fonts written with uppercase letters
### Environment

ESLint version: v10.8.0
@eslint/css version: v1.4.0
Node version: v22.22.3
npm version: v10.8.2
Operating System: StackBlitz WebContainer (linux)

stylesheet

### What did you do?

```js
// Configuration
// eslint.config.js

export default [
	{
		files: ["**/*.css"],
		plugins: { css },
		language: "css/css",
		rules: { "css/font-family-fallbacks": "error" },
	},
];
```

```css
a { font-family: Arial, Sans-Serif; }
b { font-family: Arial, SERIF; }
c { font-family: Serif; }
d { font: 16px Arial, MONOSPACE; }

/* control cases, correctly not reported */
e { font-family: Arial, sans-serif; }
f { font-family: serif; }
```

I expected no errors to be reported for any of the six declarations above.

Generic family keywords are pre-defined CSS keywords, and the CSS specification states that keywords are matched ASCII case-insensitively. From [CSS Values and Units Module Level 4, §4.1 Pre-defined Keywords](https://www.w3.org/TR/css-values-4/#keywords):

(Keywords are identifiers and are interpreted ASCII case-insensitively (i.e., [a-z] and [A-Z] are equivalent))

Generic families are keywords only when they are unquoted — [CSS Fonts Module Level 4](https://www.w3.org/TR/css-fonts-4/#generic-family-value) notes that `<generic-font-family>` keywords cannot be quoted, otherwise they are interpreted as a `<font-family-name>`. All four cases above are unquoted, so they are the keyword form.

So `Sans-Serif`, `SERIF` and `MONOSPACE` are valid generic font families and mean exactly the same thing as their lowercase forms. Only the lowercase forms are accepted right now.

### What actually happened?

The four declarations that use uppercase letters are reported, while the two lowercase control cases are not:

```
❯ npx eslint .

/home/projects/stackblitz-starters-adhqdqr2/test.css
  1:18  error  Use a generic font last                     css/font-family-fallbacks
  2:18  error  Use a generic font last                     css/font-family-fallbacks
  3:18  error  Use fallback fonts and a generic font last  css/font-family-fallbacks
  4:11  error  Use a generic font last                     css/font-family-fallbacks

✖ 4 problems (4 errors, 0 warnings)
```

The fact that only the uppercase variants are reported shows that the cause is case-sensitive matching.

https://stackblitz.com/edit/stackblitz-starters-adhqdqr2

### Participation

### AI acknowledgment

### Additional comments

Hello. I looked into where this comes from, and I would like to share what I found.

**Where the bug is**

In `src/rules/font-family-fallbacks.js`, the `genericFonts` set is defined entirely in lowercase (L20–L34), and the candidate tokens are compared against it without any case normalization:

| Line | Code |
| --- | --- |
| L101 | `!genericFonts.has(valueList.at(-1))` |
| L183 / L189 | `genericFonts.has(variableList[...])` |
| L198 | `genericFonts.has(valueArr[0].name)` |
| L261 | `genericFonts.has(fontsList.at(-1))` |
| L273 | `genericFonts.has(lastFont.name)` |
| L356 / L389 / L400 / L444 | `genericFonts.has(...)` |

What makes me think this is an oversight rather than an intentional decision is that the same file already normalizes case for the other kind of keyword it handles. CSS-wide keywords are lowercased in two places:

```js
// L43
return cssWideKeywords.has(value.trim().toLowerCase());

// L135, inside create()
sourceCode.lexer.cssWideKeywords.map(keyword => keyword.toLowerCase())
```

So CSS-wide keywords are treated case-insensitively, but generic font keywords in the very same rule are not.

**Precedent**

While searching for duplicates I found that this repository has fixed this same class of bug already:

- #222 `fix: make relative-font-units unit matching case-insensitive` 

`font-family-fallbacks` looks like the one rule that was left out. I could not find an existing issue or pull request covering this, so I am reporting it.

If this is accepted, I'd like to take it on, following the same approach as #222.

electrohyun:
#### Prerequisites checklist

#### AI acknowledgment

Bug fix.

> [!NOTE]
> This also removes some false negatives, so the rule reports more errors in the three cases listed under "Side effects" below. Per the [versioning policy](https://github.com/eslint/eslint#semantic-versioning-policy), this belongs in a minor release rather than a patch, so it may belong under `feat` rather than `fix`.

Generic font families are CSS keywords, so they are matched ASCII case-insensitively, but the rule compared candidates against a lowercase set without normalizing them. `Sans-Serif`, `SERIF` and `MONOSPACE` were reported even though they are valid.

The rule consults the generic font list in eleven places, covering the `font-family` property, the `font` shorthand, and the paths that resolve `var()` references. Ten of them are direct `genericFonts.has()` lookups. The eleventh is a substring match in the single value `font` branch, where the family name is not separated from the rest of the value.

I added an `isGenericFont()` helper that lowercases before the lookup and used it at the ten direct lookups, so the behaviour is consistent across paths rather than fixed only in the ones the issue happened to demonstrate.

Two places needed more than lowercasing, because they cannot tell a keyword from a family name on their own:

- **The single value `font` branch**, the eleventh site above. Since it scans the whole value, lowercasing alone would make `font: 16px "SERIF"` and `font: 16px MySerifFont` pass silently. It now removes quoted names first and requires a whole word match.
- **The list the `var()` path compares against.** It is built from `String` node values, which arrive with their quotes already stripped, mixed together with identifier names. The lookup itself is one of the ten, but lowercasing it alone would make `font-family: var(--f), "SANS-SERIF"` pass silently, so the list now keeps the original text.

That leaves two mechanisms for the same concept, which is worth explaining. The comparisons take two shapes. The ten direct lookups each receive a single token, so an exact match after lowercasing is enough, and that is what `isGenericFont()` does. The single value `font` branch receives the whole shorthand with the size still attached, so it has to search inside a string, and that search needs a word boundary. `genericFontPattern` is compiled once at module load rather than per call, and it carries no `g` flag, so there is no `lastIndex` state to worry about.

`docs/rules/font-family-fallbacks.md` is unchanged, since nothing in it describes this behaviour.

#### Side effects worth calling out

Telling keywords and family names apart in those two places means three kinds of input that were previously accepted are now reported. None of these relax the rule, and all of them are cases where there is genuinely no generic fallback:

- `font: 16px "serif"`. A quoted name in the single value `font` branch was being read as a generic keyword.
- `font-family: var(--f), "sans-serif"`. The same confusion through the `var()` path. These two are the distinction you asked me to cover with an invalid test case.
- `font: 16px MyserifFont`. The single value `font` branch matches by substring, so an unquoted family name that contains a generic keyword was accepted. Without a word boundary, lowercasing would also have let `MYSERIFFONT` through, so this one is not separable from the fix.

To check that nothing else moved, I swept the rule with a generated matrix of inputs covering every generic keyword through each of these branches and compared the output before and after. Every behaviour change falls into the three cases above, and there was no fourth category. (129 inputs became accepted, which is the intended fix, and 53 became reported, all of them instances of the three cases.)

#### Tests

- Valid cases for uppercase spellings, one per code path, so each of the ten direct lookups is covered.
- Invalid cases for quoted generic names, which is what you asked for: in `font-family`, in the `font` shorthand, and through a `var()` reference.
- Invalid cases for family names that contain a generic keyword as a substring, such as `MySerifFont` and `MyMonospaceFont`.

To confirm the new tests actually guard the change I ran it three ways.

Running the rule's own test file, `npx mocha tests/rules/font-family-fallbacks.test.js`:

| Source | Result |
| --- | --- |
| unchanged | 64 passing, 16 failing |
| lowercasing only, without the two special cases | 72 passing, 8 failing |
| this PR | 80 passing |

The middle row is the useful one. The quoted and substring tests are what catch a naive fix, and you can reproduce it on this branch by reverting the two special cases.

8 failing

(This is not the state of this PR. It shows what fails if the fix is done with lowercasing alone, without the two special cases described above.)

`npm test` and `npm run lint` both pass.

#### Related Issues

fixes #516

The whole word match in the single value `font` branch. It is the one place where I changed how matching works rather than only normalizing case, and it is where the third side effect comes from.

If you would prefer a narrower change, the way to narrow it is to leave that branch untouched entirely. Dropping only the word boundary is not an option, because lowercasing without it lets `MYSERIFFONT` through, which would be a regression. Leaving the branch alone would keep `font: 16px SERIF` broken while the other paths are fixed, so I would rather include it, but I can split it out if you disagree.

### DIFF
--- a/src/rules/font-family-fallbacks.js
+++ b/src/rules/font-family-fallbacks.js
@@ -33,6 +33,26 @@ const genericFonts = new Set([
 	"fangsong",
 ]);
 
+/*
+ * Matches a generic font family as a whole word, for the single value `font`
+ * branch that searches the shorthand where the family sits next to the size.
+ * A name such as `MySerifFont` must not match.
+ */
+const genericFontPattern = new RegExp(
+	String.raw`(?<![\w-])(?:${[...genericFonts].join("|")})(?![\w-])`,
+	"iu",
+);
+
+/**
+ * Check if the name is a generic font family.
+ * Generic font families are keywords, so they are matched case-insensitively.
+ * @param {string} name The name to check.
+ * @returns {boolean} True if the name is a generic font family, false otherwise.
+ */
+function isGenericFont(name) {
+	return genericFonts.has(name.toLowerCase());
+}
+
 /**
  * Check if the value is a CSS-wide keyword.
  * @param {string} value The value to check.
@@ -87,9 +107,9 @@ function reportFontWithoutFallbacksInFontProperty(
 	const valueList = fontPropertyValues.split(",").map(v => v.trim());
 
 	if (valueList.length === 1) {
-		const containsGenericFont = Array.from(genericFonts).some(font =>
-			valueList[0].includes(font),
-		);
+		// A quoted name is a family name, not a keyword
+		const unquoted = valueList[0].replace(/"[^"]*"|'[^']*'/gu, "");
+		const containsGenericFont = genericFontPattern.test(unquoted);
 
 		if (!containsGenericFont) {
 			context.report({
@@ -98,7 +118,7 @@ function reportFontWithoutFallbacksInFontProperty(
 			});
 		}
 	} else {
-		if (!genericFonts.has(valueList.at(-1))) {
+		if (!isGenericFont(valueList.at(-1))) {
 			context.report({
 				loc: node.loc,
 				messageId: "useGenericFont",
@@ -180,13 +200,13 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 
 						if (
 							variableList.length === 1 &&
-							!genericFonts.has(variableList[0])
+							!isGenericFont(variableList[0])
 						) {
 							context.report({
 								loc: node.loc,
 								messageId: "useFallbackFonts",
 							});
-						} else if (!genericFonts.has(variableList.at(-1))) {
+						} else if (!isGenericFont(variableList.at(-1))) {
 							context.report({
 								loc: node.loc,
 								messageId: "useGenericFont",
@@ -195,7 +215,7 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 					} else {
 						if (
 							valueArr[0].type === "Identifier" &&
-							genericFonts.has(valueArr[0].name)
+							isGenericFont(valueArr[0].name)
 						) {
 							return;
 						}
@@ -230,7 +250,10 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 
 						valueArr.forEach(child => {
 							if (child.type === "String") {
-								fontsList.push(child.value);
+								// Keep the quotes so it is not read as a keyword
+								fontsList.push(
+									sourceCode.getText(child).trim(),
+								);
 							}
 
 							if (child.type === "Identifier") {
@@ -258,7 +281,7 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 
 						if (
 							fontsList.length > 0 &&
-							!genericFonts.has(fontsList.at(-1))
+							!isGenericFont(fontsList.at(-1))
 						) {
 							context.report({
 								loc: node.loc,
@@ -270,7 +293,7 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 
 						if (!(
 							lastFont.type === "Identifier" &&
-							genericFonts.has(lastFont.name)
+							isGenericFont(lastFont.name)
 						)) {
 							context.report({
 								loc: node.loc,
@@ -352,9 +375,7 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 								);
 
 								if (!usingVar) {
-									if (
-										!genericFonts.has(afterOperator.at(-1))
-									) {
+									if (!isGenericFont(afterOperator.at(-1))) {
 										context.report({
 											loc: node.loc,
 											messageId: "useGenericFont",
@@ -386,9 +407,7 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 
 										if (
 											variableList.length > 0 &&
-											!genericFonts.has(
-												variableList.at(-1),
-											)
+											!isGenericFont(variableList.at(-1))
 										) {
 											context.report({
 												loc: node.loc,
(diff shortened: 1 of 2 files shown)


tests/rules/font-family-fallbacks.test.js
```diff
@@ -274,5 +287,125 @@ ruleTester.run("font-family-fallbacks", rule, {
...
+					messageId: "useFallbackFonts",
+					line: 1,
+					column: 11,
+					endLine: 1,
+					endColumn: 27,
+				},
+			],
+		},
+		{
+			code: "a { font: 16px MyMonospaceFont; }",
```
Tanujkanti4441, line 387: Can you also add a test with multiple fonts with `font` property such as:

`a { font: 16px 'Roboto', 'Open Sans', 'serif'; }`

and with `var` also? 

`:root{ --foo: 'serif'; } a { font: 16px var(--foo); }`
`:root{ --foo: 'MySerifFont'; } a { font: 16px var(--foo); }`
`:root{ --foo: 'Roboto', 'Open Sans', 'serif'; } a { font: 16px var(--foo); }`

Tanujkanti4441 (commented):
Thanks for the PR. Overall looks good; just some test-related changes are requested.

tests/rules/font-family-fallbacks.test.js
electrohyun: Hello! I've added the four test cases and rebased onto latest main. 

Since this removes false negatives (more reports in three cases), would `feat` fit better than `fix`?

Tanujkanti4441 (dismissed):
LGTM, thanks! Would like another review before merging.

tests/rules/font-family-fallbacks.test.js
Tanujkanti4441: Sure! We can go with `feat`.

src/rules/font-family-fallbacks.js
```diff
@@ -230,7 +250,10 @@ export default /** @satisfies {FontFamilyFallbacksRuleDefinition} */ ({
 
 						valueArr.forEach(child => {
 							if (child.type === "String") {
-								fontsList.push(child.value);
+								// Keep the quotes so it is not read as a keyword
+								fontsList.push(
+									sourceCode.getText(child).trim(),
```
DMartens:
```suggestion
									sourceCode.getText(child),
```
The `trim` seems to be unnecessary as the first and last characters are quotes.
If there is a case for this, please add a test case for this scenario.

electrohyun: Thanks for the review, applied. This one is inside a `String` type check, so the text always starts and ends with a quote and `trim()` never changes the value. I also checked with inputs containing extra whitespace, and there was no case to add.

(+ Pushed a formatting commit on top, since removing `.trim()` let the call fit on one line.)

DMartens (approved):
Changes LGTM, thanks.

PR #522, opened 2026-08-02
feat: check duplicate comma-separated @keyframe selectors

### LINKED ISSUE #459: Rule Change: check comma-separated keyframe selector also

no-duplicate-keyframe-selectors

Generate more warnings

A new default behavior

### Example code

```css
@keyframes a {
  0%, 50% { opacity: 0; }
  50% { opacity: 1; }
}
```

It does not report `50%` as a duplicate keyframe selector.

It will report duplicate selectors in the `@keyframe` even if it is written as comma-separated selector.

### Participation

### AI acknowledgment

### Additional comments

_No response_

Tanujkanti4441:
#### Prerequisites checklist

#### AI acknowledgment

To also check the duplicate @keyframe selectors that are comma-separated.

Added a loop to go through every selector in a rule instead of just the first one to enable checking comma-separated selectors as duplicates, and added related test cases also. Now rule `no-duplicate-keyframe-selectors` will also report cases such as:

```css
@keyframes a {
	0%, 50% { top: 0; }
	50% { top: 1px; }
}

@keyframes a {
	from, 0% { top: 0; }
	50% { top: 1px; }
}

@keyframes a {
	entry 0%, { top: 0; }
	entry 0%, 50% { top: 1px; }
}
```

#### Related Issues
Fixes #459

### DIFF
--- a/src/rules/no-duplicate-keyframe-selectors.js
+++ b/src/rules/no-duplicate-keyframe-selectors.js
@@ -62,51 +62,32 @@ export default /** @satisfies {DuplicateKeyframeSelectorRuleDefinition} */ ({
 				}
 
 				// @ts-ignore - children is a valid property for prelude
-				const selectors = node.prelude.children;
-				const value = [];
+				node.prelude.children.forEach(selector => {
+					const value = [];
 
-				selectors.forEach(selector => {
-					const component = selector.children[0];
-					const componentType = component.type;
-
-					if (selector.children.length === 1) {
-						if (componentType === "Percentage") {
+					selector.children.forEach(component => {
+						if (component.type === "Percentage") {
 							value.push(`${component.value}%`);
-						} else if (componentType === "TypeSelector") {
+						} else if (component.type === "TypeSelector") {
 							value.push(component.name.toLowerCase());
 						}
-					} else {
-						if (
-							selector.children.some(
-								child => child.type === "Combinator",
-							)
-						) {
-							if (
-								componentType === "TypeSelector" &&
-								selector.children[1].type === "Combinator" &&
-								selector.children[2].type === "Percentage"
-							) {
-								value.push(
-									`${component.name.toLowerCase()} ${selector.children[2].value}%`,
-								);
-							}
-						}
-					}
-				});
+					});
 
-				const keys = value.map(
-					selectorPart =>
-						keyframeSelectorAliases.get(selectorPart) ??
-						selectorPart,
-				);
+					const selectorValue = value.join(" ");
+					const key = value
+						.map(
+							selectorPart =>
+								keyframeSelectorAliases.get(selectorPart) ??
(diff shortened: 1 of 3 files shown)


src/rules/no-duplicate-keyframe-selectors.js
```diff
@@ -62,37 +62,57 @@ export default /** @satisfies {DuplicateKeyframeSelectorRuleDefinition} */ ({
...
+						} else if (componentType === "TypeSelector") {
+							value.push(component.name.toLowerCase());
+						}
+					} else {
+						if (
+							selector.children.some(
+								child => child.type === "Combinator",
+							)
+						) {
+							if (
```
DMartens: This would not support selectors with more than two selectors, e.g.:
```css
@keyframes multiple {
  0%, 10%, 20% {}
}
```

Tanujkanti4441: It does support multiple selectors, have added tests for this in https://github.com/eslint/css/pull/522/commits/c269dfc97bb34ed7409c49b3111728b4e9f09554. This code actually checks for a selector like `entry 0%`.

DMartens (dismissed):
Changes LGTM, thanks. As this is a feature leaving open for a second review.

src/rules/no-duplicate-keyframe-selectors.js
```diff
@@ -62,37 +62,57 @@ export default /** @satisfies {DuplicateKeyframeSelectorRuleDefinition} */ ({
 				}
 
 				// @ts-ignore - children is a valid property for prelude
-				const selector = node.prelude.children[0];
```
Pixel998, line 65: Could we simplify this change by iterating over `node.prelude.children` here and keeping the existing logic below? Is there a reason for the additional changes?
```js
node.prelude.children.forEach(selector => {
	const value = [];

	selector.children.forEach(component => {
		if (component.type === "Percentage") {
			value.push(`${component.value}%`);
		} else if (component.type === "TypeSelector") {
			value.push(component.name.toLowerCase());
		}
	});

	const selectorValue = value.join(" ");
	const key = value
		.map(
			selectorPart =>
				keyframeSelectorAliases.get(selectorPart) ??
				selectorPart,
		)
		.join(" ");

	if (seen.has(key)) {
		context.report({
			loc: selector.loc,
			messageId: "duplicateKeyframeSelector",
			data: {
				selector: selectorValue,
			},
		});
	} else {
		seen.add(key);
	}
});
```

Tanujkanti4441: Thanks for the suggestion; I have applied it. Just trying to avoid nested loop earlier.

Pixel998 (approved):
LGTM, thanks! Leaving open for @DMartens.

DMartens (approved):

PR #526, opened 2026-08-08
fix: correctly identify font-size in font shorthand

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

#### What did you do?

```css
/* eslint css/relative-font-units: "error" */

.a {
  font: oblique 10deg 1rem serif;
}

.b {
  font: 1rem math;
}
```

Neither declaration should be reported: `10deg` is the `oblique` angle, and `math` is a generic font family.

#### What actually happened?

The rule reported `10deg` and `math` instead of identifying `1rem` as the size.

Updated the rule to use the lexer’s `font` grammar match to identify the actual `font-size` component before checking whether its unit or keyword is allowed.

### DIFF
--- a/src/rules/relative-font-units.js
+++ b/src/rules/relative-font-units.js
@@ -9,6 +9,7 @@
 
 /**
  * @import { CSSRuleDefinition } from "../types.js"
+ * @import { CssNode, CssNodePlain } from "@eslint/css-tree"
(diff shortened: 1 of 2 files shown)


DMartens (approved):
Changes LGTM, thanks.

PR #530, opened 2026-08-13
fix: handle CSS case-insensitivity

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

CSS properties, keywords, functions, units, media features, and pseudo-class/element names are case-insensitive. This PR ensures that the affected rules recognize them regardless of capitalization.

Updated the affected rules to handle these CSS constructs case-insensitively.

If this is difficult to review as a single PR, I’m happy to split it into smaller PRs.

### DIFF
--- a/src/rules/prefer-logical-properties.js
+++ b/src/rules/prefer-logical-properties.js
@@ -192,6 +192,12 @@ export default /** @satisfies {PreferLogicalPropertiesRuleDefinition} */ ({
 
 	create(context) {
 		const [{ allowProperties, allowUnits }] = context.options;
+		const allowedProperties = new Set(
+			allowProperties.map(property => property.toLowerCase()),
+		);
+		const allowedUnits = new Set(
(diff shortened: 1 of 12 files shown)


src/rules/prefer-logical-properties.js
```diff
@@ -192,6 +192,12 @@ export default /** @satisfies {PreferLogicalPropertiesRuleDefinition} */ ({
 
 	create(context) {
 		const [{ allowProperties, allowUnits }] = context.options;
+		const allowedProperties = new Set(
+			allowProperties.map(property => property.toLowerCase()),
```
DMartens: This conversion to lowercase is unnecessary as `allowProperties` and `allowUnits` can only be lowercase as the provided `enum` in the schema for both is in lowercase (both maps save the properties / units in lowercase).

Pixel998: Good catch.

DMartens (approved):
Changes LGTM, thanks.

PR #535, opened 2026-08-16
feat: add no-unknown-animations rule

### LINKED ISSUE #529: New Rule: no-unknown-animations
### Rule details

I would like to propose a rule that warns when an animation name used in `animation` or `animation-name` does not have a matching `@keyframes` rule in the same source.

The rule would only check statically determinable animation names and ignore dynamic values such as `var()`.

### Example code

```css
.card {
    animation: fade-in 300ms ease;
}

.button {
    animation-name: slide-up;
}

@keyframes fade-out {
    from {
        opacity: 1;
    }

    to {
        opacity: 0;
    }
}

// In the example above, `fade-in` and `slide-up` are used as animation names, 
// but there are no matching `@keyframes` rules with those names defined in the same source,
// so I would expect the rule to report them.
```

### Prior Art

Stylelint has a `no-unknown-animations` rule with the same purpose.

The rule checks animation names used in `animation` and `animation-name`, and treats `@keyframes` names defined in the same source or in files provided through `referenceFiles` as known animations.

Documentation: https://stylelint.io/user-guide/rules/no-unknown-animations/

### Participation

### AI acknowledgment

### Additional comments

CSS can reference `@keyframes` rules defined in another stylesheet, so a missing `@keyframes` rule in the current source does not always mean the animation is undefined.

For this reason, I propose checking only animation names that can be determined statically and `@keyframes` rules defined in the same source for now.

I would appreciate feedback on whether this scope would be appropriate for `@eslint/css`, or whether cross-file cases should be handled differently.

Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

Gaic4o:
#### AI acknowledgment

This PR adds the `no-unknown-animations` rule to report animation names that don't match any `@keyframes` rule defined in the same source.

* Added the `no-unknown-animations` rule for `animation` and `animation-name` declarations.
* Added tests and documentation for the new rule.

#### Related Issues

fixes #529

Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

## Summary by CodeRabbit

- **New Features**
  - Added a CSS linting rule that reports animation names without matching `@keyframes` definitions.
  - Supports shorthand and `animation-name` declarations, vendor-prefixed properties, nested styles, and `var()` fallbacks.

- **Documentation**
  - Added usage guidance, examples, limitations, and rule-table documentation for the new rule.

- **Tests**
  - Added comprehensive coverage for valid and invalid animation names, including keyword values, quoting, casing, and nested declarations.

### DIFF
--- a/src/rules/no-unknown-animations.js
+++ b/src/rules/no-unknown-animations.js
@@ -4,6 +4,12 @@
  */
 
 //-----------------------------------------------------------------------------
+// Imports
+//-----------------------------------------------------------------------------
+
+import { parse } from "@eslint/css-tree";
+
+//-----------------------------------------------------------------------------
 // Type Definitions
 //-----------------------------------------------------------------------------
 
@@ -18,7 +24,8 @@
 // Helpers
 //-----------------------------------------------------------------------------
 
-const animationPropertyPattern = /^animation(?:-name)?$/iu;
+const animationPropertyPattern =
+	/^(?:-(?:o|moz|webkit)-)?animation(?:-name)?$/iu;
 
 /**
  * Extracts an animation name from a node. Quoted and unquoted animation
@@ -37,6 +44,81 @@ function getAnimationName(node) {
 	}
 
 	return null;
+}
+
+/**
+ * Returns the children of a node as an array. Nodes coming from the rule's
+ * AST store children in an array, while nodes produced by `parse()` store
+ * them in a list.
+ * @param {Object} node The node to read the children of.
+ * @returns {Array<Object>} The children of the node.
+ */
+function getChildren(node) {
+	const { children } = node;
+
+	if (!children) {
+		return [];
+	}
+
+	return Array.isArray(children) ? children : children.toArray();
+}
+
+/**
+ * Finds every `var()` function inside a node.
+ * @param {Object} node The node to search.
+ * @param {Array<Object>} varFunctions The array to collect the functions into.
+ * @returns {Array<Object>} The `var()` functions found.
+ */
+function findVarFunctions(node, varFunctions) {
+	for (const child of getChildren(node)) {
+		if (child.type === "Function" && child.name.toLowerCase() === "var") {
+			varFunctions.push(child);
+		}
+
+		findVarFunctions(child, varFunctions);
+	}
+
+	return varFunctions;
+}
+
+/**
+ * Replaces every `var()` with whitespace, keeping any fallback value where it
+ * was. Because the replacement is the same length as the text it replaces,
+ * the remaining value keeps the offsets it has in the original source.
+ * @param {string} text The value text to mask.
+ * @param {number} baseOffset The offset at which `text` starts in the source.
+ * @param {Array<Object>} varFunctions The `var()` functions to mask.
+ * @returns {string} The masked value text.
+ */
+function maskVarFunctions(text, baseOffset, varFunctions) {
+	/** @type {Array<[number, number]>} */
+	const ranges = [];
+
+	for (const varFunction of varFunctions) {
+		const start = varFunction.loc.start.offset - baseOffset;
+		const end = varFunction.loc.end.offset - baseOffset;
+		const fallback = getChildren(varFunction).find(
+			child => child.type === "Raw",
+		);
+
+		if (fallback) {
+			ranges.push([start, fallback.loc.start.offset - baseOffset]);
+			ranges.push([fallback.loc.end.offset - baseOffset, end]);
+		} else {
+			ranges.push([start, end]);
+		}
+	}
+
+	let masked = text;
+
+	for (const [start, end] of ranges) {
+		masked =
+			masked.slice(0, start) +
+			" ".repeat(end - start) +
+			masked.slice(end);
+	}
+
+	return masked;
 }
 
 //-----------------------------------------------------------------------------
@@ -59,7 +141,8 @@ export default /** @satisfies {NoUnknownAnimationsRuleDefinition} */ ({
 	},
 
 	create(context) {
-		const lexer = context.sourceCode.lexer;
+		const { sourceCode } = context;
+		const { lexer } = sourceCode;
 
 		/** @type {Set<string>} */
 		const definedAnimations = new Set();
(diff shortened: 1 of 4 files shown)


src/rules/no-unknown-animations.js
```diff
@@ -0,0 +1,141 @@
...
+ * @import { CssLocationRange } from "@eslint/css-tree"
+ * @typedef {"unknownAnimation"} NoUnknownAnimationsMessageIds
+ * @typedef {CSSRuleDefinition<{ RuleOptions: [], MessageIds: NoUnknownAnimationsMessageIds }>} NoUnknownAnimationsRuleDefinition
+ */
+
+//-----------------------------------------------------------------------------
+// Helpers
+//-----------------------------------------------------------------------------
+
+const animationPropertyPattern = /^animation(?:-name)?$/iu;
```
DMartens: This should also check for vendored prefixes (e.g. `-webkit-animation`) as the `@keyframes` check also does.

src/rules/no-unknown-animations.js
```diff
@@ -0,0 +1,141 @@
...
+
+				const matchResult = lexer.matchProperty(
+					node.property,
+					node.value,
+				);
+
+				/*
+				 * If the value can't be matched against the property grammar,
+				 * its animation name can't be determined reliably. This
+				 * includes dynamic values such as var(). Invalid property
```
DMartens: I think the rule should support checking local resolvable `var` declarations.
There will be a helper for this but this rule could already check the default value of a `var`, e.g. "slide-in" in `animation: var(--animation-name, "slide-in")`.

src/rules/no-unknown-animations.js
```diff
@@ -0,0 +1,141 @@
...
+				if (matchResult.error) {
+					return;
+				}
+
+				for (const child of node.value.children) {
+					if (!matchResult.isType(child, "keyframes-name")) {
+						continue;
+					}
+
+					const name = getAnimationName(child);
```
DMartens: The `name` should not be null as the lexer already checks that it is a string or an identifier. Otherwise a test case for this is missing.

src/rules/no-unknown-animations.js
Gaic4o: Thanks for pointing this out. After looking into it, I think it makes more sense for this rule to check values that can be determined statically, rather than trying to fully resolve every `var()` usage.

For example, `var(--animation-name)` would still be ignored when its actual value cannot be determined, while cases such as `var(--animation-name, "slide-in")` could be checked by extracting the statically known animation name from the fallback value.

For resolving local custom property values themselves, I think it would be better not to implement that separately in this PR, and instead make use of the helper you mentioned once it is available. So for this PR, I’m planning to support checking statically resolvable fallback values first.

Gaic4o: I ended up implementing `var()` handling a little more broadly than I initially described. Even when a value contains `var()`, the rule now checks fallback values as well as any statically known animation names around it. Actual custom property value resolution is still something I plan to handle later using the helper you mentioned.

One thing I’d like your opinion on is that this implementation re-parses the value using `parse()` from `@eslint/css-tree`. Since this does not use the custom parser when `customSyntax` is configured, I’d like to know whether you think this approach is okay.

src/rules/no-unknown-animations.js
Gaic4o: I removed the `null` check on the usage side. As you pointed out, the lexer only matches an identifier or a string as `<keyframes-name>`, so it can't be `null` in this case.

I kept the check on the `@keyframes` prelude side, though. This part isn't validated by the lexer, so `@keyframes 50%` can be parsed as `Percentage` and `@keyframes 1s` as `Dimension`. I also added tests for both cases.

src/rules/no-unknown-animations.js
Gaic4o: I’ve addressed the issue you pointed out. Thank you!

Gaic4o:
@DMartens  I’ve updated the PR based on the previous feedback. When you have a chance, I’d appreciate it if you could take another look.

src/rules/no-unknown-animations.js
```diff
@@ -59,19 +141,99 @@ export default /** @satisfies {NoUnknownAnimationsRuleDefinition} */ ({
...
+					name: /** @type {string} */ (getAnimationName(child)),
+					loc: child.loc,
+				});
+			}
+
+			return names;
+		}
+
 		return {
 			"Atrule[name=/^(-(o|moz|webkit)-)?keyframes$/i] > AtrulePrelude"(
```
DMartens, line 228:
```suggestion
			"Atrule[name=/^(-(o|ms|moz|webkit)-)?keyframes$/i] > AtrulePrelude"(
```
The "ms" vendor-prefix is also missing here. Please also add a test case for this.

src/rules/no-unknown-animations.js
```diff
@@ -39,6 +46,81 @@ function getAnimationName(node) {
...
+/**
+ * Replaces every `var()` with whitespace, keeping any fallback value where it
+ * was. Because the replacement is the same length as the text it replaces,
+ * the remaining value keeps the offsets it has in the original source.
+ * @param {string} text The value text to mask.
+ * @param {number} baseOffset The offset at which `text` starts in the source.
+ * @param {Array<Object>} varFunctions The `var()` functions to mask.
+ * @returns {string} The masked value text.
+ */
+function maskVarFunctions(text, baseOffset, varFunctions) {
```
DMartens, line 93: Rather than using this hack, we should not use the `lexer` to find the animation names (it is okay that the property values may be wrong).

src/rules/no-unknown-animations.js
```diff
@@ -39,6 +46,81 @@ function getAnimationName(node) {
...
 }
 
+/**
+ * Returns the children of a node as an array. Nodes coming from the rule's
+ * AST store children in an array, while nodes produced by `parse()` store
+ * them in a list.
+ * @param {Object} node The node to read the children of.
+ * @returns {Array<Object>} The children of the node.
+ */
+function getChildren(node) {
```
DMartens, line 56: The parser already converts every csstree list to an array.
As such this function is unnecessary and could be replaced with `node.children ?? []` as it checks whether the node has children (e.g. may be called with an `Identifier`)

src/rules/no-unknown-animations.js
```diff
@@ -18,7 +24,8 @@
 // Helpers
 //-----------------------------------------------------------------------------
 
-const animationPropertyPattern = /^animation(?:-name)?$/iu;
+const animationPropertyPattern =
+	/^(?:-(?:o|moz|webkit)-)?animation(?:-name)?$/iu;
```
DMartens, line 28:
```suggestion
	/^(?:-(?:o|moz|ms|webkit)-)?animation(?:-name)?$/iu;
```
The "ms" vendor-prefix is missing. Please also add a test case for this.

DMartens (changes requested):
Sorry for the delay.

src/rules/no-unknown-animations.js
Gaic4o: Agreed, the `ms` prefix was missing. I added `ms` support for both `keyframes` and `animation` properties, along with related test cases. Thanks for the review!

src/rules/no-unknown-animations.js
Gaic4o: I simplified the implementation so we no longer validate the entire property value with the lexer just to find animation names. It now identifies statically known animation names directly from the existing AST and only handles `var()` fallbacks when needed. I also added related tests.

src/rules/no-unknown-animations.js
Gaic4o: Yes, thanks for pointing that out!

src/rules/no-unknown-animations.js
Gaic4o: Yes, I added the `ms` vendor prefix and a test case for it. Thanks!

src/rules/no-unknown-animations.js
```diff
@@ -0,0 +1,258 @@
...
+// Helpers
+//-----------------------------------------------------------------------------
+
+const animationPropertyPattern =
+	/^(?:-(?:o|ms|moz|webkit)-)?animation(?:-name)?$/iu;
+
+/**
+ * Keywords that `animation-name` accepts in place of an animation name.
+ */
+const animationNameKeywords = new Set([
```
DMartens, line 33: These are not "animationNameKeywords" but CSS-wide keywords.
You should access them via `sourceCode.lexer.cssWideKeywords` as the user can expand them.

src/rules/no-unknown-animations.js
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * Parses the fallback of a `var()` function, which the AST keeps as raw
+ * text. Passing the fallback's position in the source along lets the names
+ * inside it be reported where they appear.
+ * @param {Object} fallback The `Raw` node holding the fallback.
+ * @returns {Object|null} The parsed value node, or `null` if the fallback
+ *      can't be parsed.
+ */
+function parseVarFallback(fallback) {
```
DMartens, line 102: There should be no need to parse the fallback value.
You can just wrap it as a string (and remove quotes around the value):
```javascript
({ type: 'String', value: fallback.value })
```
This allows reusing the rest of the code and using a string ensures it cannot be detected as a CSS-wide keyword.

src/rules/no-unknown-animations.js
```diff
@@ -0,0 +1,258 @@
...
+ */
+function findAnimationNames(value, isShorthand, names) {
+	const keywords = isShorthand
+		? animationShorthandKeywords
+		: animationNameKeywords;
+
+	for (const child of value.children ?? []) {
+		if (child.type === "Function") {
+			if (child.name.toLowerCase() === "var") {
+				const fallback = child.children.find(
```
DMartens, line 144: Rather than checking for the "Raw" node, always use third child if it exists.
This should be more future-proof as we may enable [`parseCustomProperty`](https://github.com/csstree/csstree/blob/master/docs/parsing.md#parsecustomproperty) in the future or a user uses the "tolerant" parsing.

DMartens (changes requested):
Thank you for applying the requested changes.
I have some more notes for the newly added code.

PR #541, opened 2026-08-24
fix: crash in `no-important` on non-standard important values

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

#### What did you do?

```css
/* eslint css/no-important: "error" */
a { color: red !ie; }
```

No error. `!ie` is a non-standard hack, not an `!important` flag.

#### What actually happened?

```
TypeError: Cannot read properties of null (reading 'index')
```

This PR fixes the crash above. `important` holds the raw identifier for hacks like `!ie`, which is truthy, so the rule went on to locate the flag in the source text, found no match, and dereferenced `null`.

Narrowed the check to `important === true`, so only real `!important` flags are reported.

Depends on eslint/csstree#152, which makes `!IMPORTANT` parse as `true`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected priority detection so only declarations explicitly marked as important are reported.
  * Legacy Internet Explorer priority annotations are now accepted without triggering violations.

* **Tests**
  * Added coverage for supported `!ie`, `!IE7`, and `!important\9` annotations.


DMartens (dismissed):
Changes make sense to me, thanks. Marking as approved waiting for the csstree PR to land.

DMartens:
Can you please rebase on main as there was a `@eslint/csstree` release with the required fix?

DMartens (approved):
Changes LGTM, thanks.

PR #542, opened 2026-08-25
feat: check CSS-wide keywords in `use-baseline`

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

#### What did you do?

```css
/* eslint css/use-baseline: ["error", { available: 2021 }] */

a {
  font-size: revert-layer;
}
```

The declaration to be reported. [`revert-layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/revert-layer) only reached Baseline in 2022, so with `{ available: 2021 }` it is not yet available.

#### What actually happened?

No error is reported.

This PR makes `use-baseline` check CSS-wide keywords.

- Added a `globalKeywords` map to the generated baseline data, sourced from the `css.types.global_keywords.*` compat keys.
- `checkPropertyValueIdentifier` now consults that map before the per-property `propertyValues` lookup.

#### Additional comments

[`revert-rule`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/revert-rule) is not covered. `web-features` does not expose it yet, so there is no Baseline status to check against. The pattern added here already matches its compat key, so it will be picked up automatically.

### DIFF
--- a/src/data/baseline-data.js
+++ b/src/data/baseline-data.js
@@ -900,6 +900,13 @@ export const selectors = new Map([
 	["xr-overlay", "0:"],
 	["where", "10:2021"],
 ]);
+export const globalKeywords = new Map([
+	["revert-layer", "10:2022"],
+	["inherit", "10:2015"],
+	["initial", "10:2015"],
+	["revert", "10:2020"],
+	["unset", "10:2016"],
+]);
 export const propertyValues = new Map([
 	[
 		"position",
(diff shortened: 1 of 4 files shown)


nzakas (approved):
LGTM. Just waiting another day to give others time to review.

DMartens (approved):
Changes LGTM, thanks. Leaving open for a second review.
Edit: As it has two reviews, moving to merge candidates.

lumirlumir (approved):
LGTM, thanks!

It’s been four more days, so I’m merging it.

PR #550, opened 2026-08-30
fix: locate `!important` after the value in `no-important`

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

#### What did you do?

```css
/* eslint css/no-important: "error" */

a { content: "!important" !important; }
```

The report to point at the actual `!important` flag.

#### What actually happened?

The rule matched the first `!important` in the declaration text, which is the one inside the string.

This PR fixes `no-important` reporting the wrong location.

The flag can only appear after the value, so the search now starts at `node.value.loc.end.offset` instead of the start of the declaration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of `!important` detection in CSS declarations.
  * Prevented false positives for `!important` text in property names, quoted strings, URLs, and custom property values.
  * Ensured actual trailing `!important` declarations continue to be reported and removed correctly.

* **Tests**
  * Added coverage for quoted strings, URLs, and custom properties containing `!important`.

### DIFF
--- a/src/rules/no-important.js
+++ b/src/rules/no-important.js
@@ -58,12 +58,16 @@ export default /** @satisfies {NoImportantRuleDefinition} */ ({
 						/* eslint-disable-next-line require-unicode-regexp -- we want to replace each code unit with a space */
 						match => match.replace(/[^\r\n\f]/g, " "),
 					);
-					const importantMatch =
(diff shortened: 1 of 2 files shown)


DMartens (approved):
Changes LGTM, thanks.

PR #564, opened 2026-09-13
fix: update baseline data

eslintbot:
Updates baseline data using the latest versions of @webref/css and web-features.

This PR is autogenerated by the "Update Baseline" GitHub Action.

## Summary by CodeRabbit

- **Updates**
  - Updated web platform compatibility data with additional support information for link parameters and parameter functions.
  - Refined browser support baselines for several CSS features, including `rx`, `ry`, `text-wrap`, `url`, and `text-decoration-thickness`.
- **Chores**
  - Updated the web features data package used to maintain compatibility information.

PR #566, opened 2026-09-14
fix: update baseline data

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

Updates baseline data to `web-features@3.38.0` and fixes the generator so `url()` keeps its widely available status.

web-features 3.38.0 adds the `link-parameters` feature with the compat key `css.types.param.url`. The generator read it as the `url()` function and overwrote the status from `css.types.url`.

- Bumped `web-features` to `^3.38.0` and regenerated the baseline data.
- A function's own `css.types.<name>` key now takes precedence over `css.types.<group>.<name>` keys and feature-name matches.
- Added `shape` to the list of bare type keys that aren't functions, since `css.types.shape` is the deprecated `<shape>` type, not `shape()`.

#### Related Issues

Closes #564

## Summary by CodeRabbit

- **Compatibility Updates**
  - Updated browser support information for several CSS features, including `rx`, `ry`, `text-wrap`, and percentage-based `text-decoration-thickness`.
  - Added compatibility data for the `link-parameters` property and `param()` function.
  - Improved compatibility reporting for CSS functions, including more accurate handling of `shape()`.


tools/generate-baseline.js
```diff
@@ -258,7 +278,12 @@ function extractCSSFeatures(features) {
 			else if ((match = PATTERNS.type.exec(key))) {
 				const { group, type } = match.groups;
 				if (isCompatTypeAFunction(group, type)) {
-					output.functions[type] = mapFeatureStatus(status);
+					if (!group) {
```
DMartens, line 281: Maybe this logic would be easier to understand like this:
```javascript
if (!functionsFromBareKeys.has(type)) {
	output.functions[type] = mapFeatureStatus(status);
}

if (!group) {
	functionsFromBareKeys.add(type);
}
```
The logic should be equivalent (currently produces the same output).

Pixel998: Updated in https://github.com/eslint/css/pull/566/commits/998c09825ec36dd8a9eea3cf5a614c5ee3ec1f4b.

DMartens (approved):
Changes LGTM, thank you.

PR #532, opened 2026-08-13
test: cover `@supports` feature functions

Pixel998:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

#### AI acknowledgment

- [x] I did _not_ use AI to generate this PR.
- [ ] (If the above is not checked) I have reviewed the AI-generated content before submitting.

#### What is the purpose of this pull request?

This PR adds test coverage for `@supports` feature functions and nested conditions.

#### What changes did you make? (Give an overview)

Added `no-invalid-at-rules` test cases for nested conditions and all supported `@supports` feature functions.

#### Related Issues

Blocked on eslint/csstree#148

PR #570, opened 2026-09-19
fix: handle `var()` and `env()` in `no-invalid-at-rules`

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

#### What did you do?

```css
/* eslint css/no-invalid-at-rules: "error" */

@supports (padding: env(safe-area-inset-top)) {
	body {
		padding-top: env(safe-area-inset-top);
	}
}
```

No lint error. The prelude is valid, and the lexer does not support matching values containing `env()` or `var()`, so the rule should skip them, as `no-invalid-properties` already does for `env()`.

#### What actually happened?

The rule crashed:

```
TypeError: Cannot read properties of null (reading '1')
```

This PR makes `no-invalid-at-rules` skip preludes and descriptor values containing `var()` or `env()`, so they no longer crash the rule or get reported as unknown descriptors.

- `no-invalid-at-rules` now returns early when the lexer reports it cannot match a prelude or descriptor value containing `var()` or `env()`.
- Replaced `isEnvMatchError`, which compared CSSTree's message text and only covered `env()`, with `isUnsupportedMatchingTreeError`, which checks the error's `name` and covers both functions. `no-invalid-properties` now uses it.

#### Related Issues

refs #532 (its `@supports env(safe-area-inset-top) {}` test case crashes without this fix)

## Summary by CodeRabbit

- **Bug Fixes**
  - CSS validation now correctly handles unresolved `var()` and user-agent `env()` values in at-rule conditions and descriptors without reporting false errors.
  - Genuine syntax and other lexer errors continue to be reported as before.

- **Tests**
  - Added coverage for `var()` and `env()` usage in `@supports` conditions and `@page` descriptors.


DMartens (approved):
Changes LGTM, thanks. Leaving open for others to review.

PR #571, opened 2026-09-19
fix: preserve keyframe selector casing in error messages

### LINKED ISSUE #562: Change Request: Preserve the original casing of selectors in `no-duplicate-keyframe-selectors` error messages
### Environment

ESLint version: 10.8.0
@eslint/css version: ^2.0.0
Node version: v22.18.0
npm version: v10.9.3
Operating System: darwin 24.6.0

### What problem do you want to solve?

- `no-duplicate-keyframe-selectors` correctly detects duplicate selectors regardless of casing, but its error messages display the selector in lowercase rather than as written.

```css
/* eslint css/no-duplicate-keyframe-selectors: "error" */

@keyframes test {
  from {
    opacity: 0;
  }
  to {
    opacity: 1;
  }
  TO {
    opacity: 2;
  }
}
```

- The current error message is:
```
Unexpected duplicate selector 'to' found within keyframe block.
```

- The error location points to `TO`, but the message displays to. 
- Preserving the reported selector’s original casing could make it easier to match the message to the code.
- The same plugin’s `prefer-logical-properties` rule preserves the original casing in the following examples: [StackBlitz reproduction](https://stackblitz.com/edit/vitejs-vite-18mmtr5a?file=package.json,test.css,eslint.config.js,.vscode%2Fsettings.json,integrations.vscode.userData.settingsRaw).
```css
/* eslint css/prefer-logical-properties: "error" */

.example {
  MARGIN-LEFT: 16PX;
  text-align: RIGHT;
}
```
```
Expected logical property 'margin-inline-start' instead of 'MARGIN-LEFT'.
Expected logical value 'end' instead of 'RIGHT'.
```

### What do you think is the correct solution?

- I’d suggest keeping the value used for duplicate detection separate from the value displayed in the error message, so the message preserves the reported selector’s original casing.
- The expected message for the example above would be:
```
Unexpected duplicate selector 'TO' found within keyframe block.
```

### Participation

- [x] I am willing to submit a pull request for this change.

### AI acknowledgment

### Additional comments

> Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

KumJungMin:
#### Prerequisites checklist

#### AI acknowledgment

- Preserve the original casing of selectors in `no-duplicate-keyframe-selectors` error messages. 
- For example, a duplicate `TO` is reported as `TO` instead of `to`.

- Separate the values used for duplicate detection from those displayed in error messages.
- Update six existing test expectations to reflect the reported selector's original casing.
- Duplicate detection, error locations, and whitespace and comment handling remain unchanged.

#### Related Issues

fixes #562

<br/>

> Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

## Summary by CodeRabbit

- **Bug Fixes**
  - Duplicate keyframe selector errors now display selectors with their original type-selector casing.
  - Selector comparison remains case-insensitive, while percentage formatting stays normalized with `%`.
- **Tests**
  - Updated validation coverage to confirm duplicate selectors are reported using their source spelling.


src/rules/no-duplicate-keyframe-selectors.js
```diff
@@ -66,16 +66,20 @@ export default /** @satisfies {DuplicateKeyframeSelectorRuleDefinition} */ ({
 				// @ts-ignore - children is a valid property for prelude
 				node.prelude.children.forEach(selector => {
 					const value = [];
+					const rawValue = [];
```
DMartens: I do not think we need the variable `rawValue`.
We can just push the the values as we do currently and use the lowercase variant when computing `key` in the `map` callback (line 78).

KumJungMin: Thank you for the comment.
I hadn’t considered that keeping a separate `rawValue` array would mean maintaining the same data in two places.
I’ve updated the code as follows :) 

https://github.com/eslint/css/pull/571/changes/963d3dcbe0d87cbcd7a7cdbb773a77aeb9bf0f11

DMartens (approved):
Changes LGTM, thanks.

